PR #1, opened 2020-04-28
Setup repository

Antoninbln:


Olivier-Daire:
@Antoninbln Could you try to rebase and rerun the CI to see if coverage works now that the repo si public? 🙏  (Can't seem to find the re-run button 😞 )

.github/workflows/lint-and-test.yaml
```diff
@@ -0,0 +1,37 @@
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Lint and test coverage
+
+on: [push, pull_request]
```
benard-g, line 6: I would suggest `push` only 🤔 (to not have 2 builds on pull requests)

Antoninbln:
I'm closing the PR for the moment, because there is a missing token used in CI.
I'll reopen it if it still fits to the techs choices we've chosen, what do you think ?

Olivier-Daire:
> I'll reopen it if it still fits to the techs choices we've chosen, what do you think ?

I think that since we are now public we do not need a token anymore.
Since we chose Python I also think that this "boilerplate" could be useful. We will probably need Flask afterwards, but this already adds the Github Actions workflow and Pylint / Pytest.
Let's see what the others think

.github/workflows/lint-and-test.yaml
```diff
@@ -0,0 +1,37 @@
...
+    - name: Lint
+      run: pipenv run make lint
+    - name: Type checking
+      run: pipenv run make mypy
+    - name: Tests
+      run: pipenv run make pytest
+    - name: Coverage check
+      uses: codecov/codecov-action@v1
+      with:
+        token: ${{ secrets.CODECOV_TOKEN }}
```
maximejimenez, line 33: Remove this line since this is only required for **private** repository.
Since the repo is public we don't need it and it is working fine

README.md
```diff
@@ -1,2 +1,11 @@
 # nestor-api
-Nestor API
+
+Nestor-api is an API used to manage deployment of applications Kubernetes. I can be used with [nestor](https://github.com/ChauffeurPrive/nestor), a CLI tool consuming nestor-api.
```
maximejimenez, line 3:
```suggestion
Nestor-api is an API used to manage deployment of applications Kubernetes. I can be used with [nestor](https://github.com/ChauffeurPrive/nestor-cli), a CLI tool consuming nestor-api.
```

Makefile
```diff
@@ -0,0 +1,38 @@
+.PHONY: format lint coverage isort-format black-format
```
benard-g, line 1: Some of the rules are not listed there (and I don't see `coverage`). Is it what is expected ?

Also, instead of having a single .PHONY with all the rules, I think it is easier to prefix each rule with its own .PHONY to keep track of which are used and which are not 🙂 .

```Makefile
.PHONY: rule1
rule1:
    echo rule1

.PHONY: rule2
rule2:
    echo rule2
```

maximejimenez: Having one PHONY per rule seems a bit overkill to me as it will kind of make the file becomes very hard to read...

Pipfile
```diff
@@ -0,0 +1,21 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+setuptools = "*"
+pylint = "*"
```
Olivier-Daire, line 8: Is there any chance we could pin those dependencies? Pytest too

README.md
```diff
@@ -1,2 +1,11 @@
 # nestor-api
-Nestor API
+
+Nestor-api is an API used to manage deployment of applications Kubernetes. I can be used with [nestor-cli](https://github.com/ChauffeurPrive/nestor-cli), a CLI tool consuming nestor-api.
+
+## Installing
+
+Dependencies are managed by [Pipenv](https://github.com/pypa/pipenv).
```
Olivier-Daire, line 7: We might also want to put at least the Python version required

nestor/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""nestor-api main module."""
+
+__author__ = "Kapten"
+__email__ = "tech-admin@kapten.com"
```
Olivier-Daire, line 4: Maybe we could put `tech@kapten.com` just like in `setup.py` (I am not sure we will be able to keep this address for long, though

README.md
benard-g: Going even further : if we want to support multiple python version we could have the CI check on each version.
https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#starting-with-the-python-workflow-template
(Since the setup says "at least 3.6")

nestor/__init__.py
maximejimenez: I'll just set `tech-admin` everywhere.

.isort.cfg
```diff
@@ -0,0 +1,7 @@
+[settings]
+multi_line_output=3
+include_trailing_comma=True
+force_grid_wrap=0
+force_sort_within_sections=True
+use_parentheses=True
+line_length=80
```
Olivier-Daire, line 7: Small detail, but I think we could allow `100` , `80` is a bit extreme considering today's screen size

Olivier-Daire: Plus it would be the same as what is definer in `.pylintrc`

Makefile
```diff
@@ -0,0 +1,38 @@
...
+	@echo "isort-format - Format python code with isort"
+	@echo "lint - Check style with pylint"
+	@echo "mypy - Run the static type checker"
+	@echo "pytest - Run tests suite with pytest"
+	@echo "test-all - Run format, lint, and test coverage"
+	@echo "vulncheck - Check for packages vulnerabilities with pipenv"
+
+black-format:
+	@echo "$(BOLD)Formatting using black$(ENDBOLD)"
+	black --line-length 80 nestor ./**.py
```
benard-g, line 18:
```suggestion
	black --line-length 100 nestor ./**.py
```

benard-g:
You could also add the CI badge 🙂  (not sure which PR will be merged first in `master`)

[Lint and test coverage](https://github.com/ChauffeurPrive/nestor-api/actions?query=branch%3Amaster)

PR #6, opened 2020-06-16
Add web server

maximejimenez:
Create web server with flask.
There is a public route (heartbeat) and a sample route (Hello world).
There is currently no authentication, no metrics, no logger embedded.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,17 @@
+__pycache__
+*.pyc
+*.pyo
+*.pyd
+.Python
+env
+pip-log.txt
+pip-delete-this-directory.txt
+.tox
+.coverage
+.coverage.*
+.cache
+nosetests.xml
+coverage.xml
+*,cover
+*.log
+.git
--- a/nestor_api/api/api_routes/sample.py
+++ b/nestor_api/api/api_routes/sample.py
@@ -3,7 +3,7 @@ from flask import Blueprint, Response, jsonify
 
 
 def register_routes(api: Blueprint) -> None:
-    """Register the /sample route."""
+    """Register the /sample routes."""
 
     @api.route("/sample", methods=["GET"])
     def _sample() -> Response:
--- a/nestor_api/api/api_routes/sample.py
+++ b/nestor_api/api/api_routes/sample.py
@@ -6,6 +6,6 @@ def register_routes(api: Blueprint) -> None:
     """Register the /sample route."""
 
     @api.route("/sample", methods=["GET"])
-    def sample() -> Response:  # pylint: disable=unused-variable
+    def _sample() -> Response:
         """Return a hello world."""
         return jsonify({"sample": "Hello, world!"})
--- a/nestor_api/api/api_routes/sample.py
+++ b/nestor_api/api/api_routes/sample.py
@@ -1,0 +1,11 @@
+"""Define the sample route."""
+from flask import Blueprint, Response, jsonify
+
+
+def register_routes(api: Blueprint) -> None:
+    """Register the /sample route."""
+
+    @api.route("/sample", methods=["GET"])
+    def sample() -> Response:  # pylint: disable=unused-variable
+        """Return a hello world."""
+        return jsonify({"sample": "Hello, world!"})
--- a/nestor_api/api/flask_app.py
+++ b/nestor_api/api/flask_app.py
@@ -1,4 +1,4 @@
-"""Return a initialized Flask application with the API."""
+"""Return an initialized Flask application with the API."""
 from flask import Flask
 
 from nestor_api.api.api import create_api
--- a/nestor_api/api/flask_app.py
+++ b/nestor_api/api/flask_app.py
@@ -1,0 +1,18 @@
+"""Return a initialized Flask application with the API."""
+from flask import Flask
+
+from nestor_api.api.api import create_api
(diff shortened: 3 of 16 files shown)


nestor_api/api/api_routes/sample.py
```diff
@@ -0,0 +1,11 @@
+"""Define the sample route."""
+from flask import Blueprint, Response, jsonify
+
+
+def register_routes(api: Blueprint) -> None:
+    """Register the /sample route."""
+
+    @api.route("/sample", methods=["GET"])
+    def sample() -> Response:  # pylint: disable=unused-variable
```
benard-g, line 9: It may be good to fix this warning as I am afraid we will have it for every route we add.
Try having a look at this --> https://github.com/jschaf/pylint-flask
(disclaimer: this is the first result I found, I did not try it 🙂 )

tests/api/api_routes/test_sample.py
```diff
@@ -0,0 +1,13 @@
+"""Test nestor_api.api.api_routes.sample"""
```
benard-g, line 1:
```suggestion
"""Test nestor_api.api.api_routes.sample."""
```

nestor_api/api/flask_app.py
```diff
@@ -0,0 +1,18 @@
+"""Return a initialized Flask application with the API."""
```
chris-verclytte, line 1:
```suggestion
"""Return an initialized Flask application with the API."""
```

nestor_api/api/api_routes/sample.py
```diff
@@ -0,0 +1,11 @@
+"""Define the sample route."""
+from flask import Blueprint, Response, jsonify
+
+
+def register_routes(api: Blueprint) -> None:
+    """Register the /sample route."""
+
+    @api.route("/sample", methods=["GET"])
```
chris-verclytte, line 8: Do we really need this sample that will be removed quite soon anyway ?

nestor_api/api/public_routes/heartbeat.py
```diff
@@ -0,0 +1,10 @@
+"""Define the heartbeat route."""
+from flask import Blueprint, jsonify
+
+blueprint = Blueprint("heartbeat", __name__)
+
+
+@blueprint.route("/heartbeat")
+def heartbeat():
+    """A heartbeat route for monitoring purpose (no auth!)."""
+    return jsonify({"state": "up"})
```
chris-verclytte, line 10: I think we should rather just respond with a 200 without body.

nestor_api/api/flask_app.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+def create_app() -> Flask:
+    """Initialize a Flask app."""
+    app = Flask(__name__)
+
+    # Note: heartbeat is exposed without authentication
+    app.register_blueprint(heartbeat.blueprint)
+
+    api = create_api()
```
chris-verclytte, line 15: Why do we use this function rather than creating blueprint directly inside api module as it is done for heartbeat ?

nestor_api/api/api_routes/sample.py
maximejimenez: Workaroudn seems to be to prefix with `_`

nestor_api/api/api_routes/sample.py
maximejimenez: It's just to have non empty folder, but not mandatory, it is so that people don't define routes in `public_routes` which is only used for the heartbeat route.

.dockerignore
```diff
@@ -0,0 +1,17 @@
...
+pip-delete-this-directory.txt
+.tox
+.coverage
+.coverage.*
+.cache
+nosetests.xml
+coverage.xml
+*,cover
+*.log
+.git
```
benard-g, line 17: vscode: `insert final newline` 🙂

nestor_api/api/public_routes/heartbeat.py
maximejimenez: Then a 204, right?

nestor_api/api/flask_app.py
maximejimenez: Because there is no function such as `register_blueprint` on a blueprint. It does not allow us to have easily a "root" blueprint with a `/api` prefix that would register a workflows (for instance) route under `/api` giving us a `/api/workflows`.

tests/api/public_routes/test_heartbeat.py
```diff
@@ -0,0 +1,13 @@
+"""Test nestor_api.api.public_routes.heartbeat."""
+from flask.testing import FlaskClient
```
benard-g, line 2: Is this import necessary ?

maximejimenez: Changed so that pylint also checks on the test folder

nestor_api/api/api_routes/sample.py
```diff
@@ -0,0 +1,11 @@
+"""Define the sample route."""
+from flask import Blueprint, Response, jsonify
+
+
+def register_routes(api: Blueprint) -> None:
+    """Register the /sample route."""
```
chris-verclytte, line 6:
```suggestion
    """Register the /sample routes."""
```
The register_routes applies for all sample routes

PR #8, opened 2020-06-18
TECH Fix warning from "itsdangerous" dependency

benard-g:
When running pytest, we end up with some warnings due to flask's dependency "itsdangerous". I tried to dump it to the LTS version and it seems to fix the issue.

I am not 100% sure about the consequences tho 😅. What do you think ?

| Before | Now |
| --- | --- |
| image | image|

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -14,6 +14,7 @@ mypy = "==0.770"
 
 [packages]
 flask = "==1.1.2"
+itsdangerous = "==1.1.0"
 
 [requires]
 python_version = "3.7"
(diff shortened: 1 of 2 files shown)


Pipfile
```diff
@@ -14,6 +14,7 @@ mypy = "==0.770"
 
 [packages]
 flask = "==1.1.2"
+itsdangerous = "==1.1.0"
```
Olivier-Daire, line 17: Maybe just add a comment saying that it is a dependency of flask and should be remove with the next flask version

PR #9, opened 2020-06-24
Add I/O library

maximejimenez:
Add a I/O library to wrap and abstract low-level interactions with system and file system.

### DIFF
--- a/nestor_api/config/config.py
+++ b/nestor_api/config/config.py
@@ -1,0 +1,10 @@
+"""Nestor-api configuration"""
+
+import os
+
+
+class Configuration: # pylint: disable=too-few-public-methods
+    """Nestor-api configuration"""
+    config_path = os.getenv("NESTOR_CONFIG_PATH")
+    pristine_path = os.getenv("NESTOR_PRISTINE_PATH", "/tmp/nestor/pristine")
+    working_path = os.getenv("NESTOR_WORK_PATH", "/tmp/nestor/work")
--- a/nestor_api/lib/io.py
+++ b/nestor_api/lib/io.py
@@ -1,0 +1,102 @@
+"""I/O library"""
+
+import errno
+import math
+import os
+import shutil
+import subprocess
+
+from datetime import datetime
+from pathlib import Path
+from random import random
+
+import yaml
+
+from nestor_api.config.config import Configuration
+
+
+def copy(source, destination):
+    """Copy file or directory with its content from source to destination"""
+    try:
+        shutil.copytree(source, destination)
+    except OSError as err:
+        if err.errno == errno.ENOTDIR:
+            shutil.copy(source, destination)
+        else:
+            raise
+
+
+def create_temporary_directory(directory_prefix=""):
+    """Creates a temporary directory with an optional prefix"""
+    tmp_directory_path = get_temporary_directory_path(directory_prefix)
+
+    # Create directory if it does not exist
+    ensure_dir(tmp_directory_path)
+
+    return tmp_directory_path
+
+
+def ensure_dir(directory_path):
+    """Ensures that a directory exists, else creates it"""
+    Path(directory_path).mkdir(parents=True, exist_ok=True)
+
+
+def execute(command: str, cwd=None):
+    """Executes a command and returns the stdout from it"""
+    result = subprocess.run(command.split(), stdout=subprocess.PIPE, check=True, cwd=cwd)
+    return result.stdout.decode('utf-8')
+
+
+def exists(file_path):
+    """Checks if a file exists"""
+    return Path(file_path).exists()
+
+
+def from_yaml(file_path):
+    """Returns a dictionary from yaml file path"""
+    with open(file_path, 'r') as file_data:
+        yaml_data = file_data.read()
+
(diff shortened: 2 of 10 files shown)


tests/lib/test_io.py
```diff
@@ -0,0 +1,203 @@
...
+        fixtures_directory_path = Path(os.path.dirname(__file__), '..', '__abcdefg').resolve()
+        destination_path = Path(os.path.dirname(__file__), '..', '__abcdefg_copy').resolve()
+
+        io.copy(fixtures_directory_path, destination_path)
+
+
+def test_create_temporary_directory():
+    temporary_path = io.create_temporary_directory("prefix")
+
+    assert temporary_path.startswith('/tmp/nestor/work/prefix-')
```
Olivier-Daire, line 52: Wondering if we should not compare to config variable here.
It is nice to have a plain string in tests, but if, for any case, the `NESTOR_WORK_PATH` variable is set, this test will fail, what do you think?

Olivier-Daire: Or maybe overrides values for the whole test file

nestor_api/lib/io.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def ensure_dir(directory_path):
+    """Ensures that a directory exists, else creates it"""
+    Path(directory_path).mkdir(mode=0o777, parents=True, exist_ok=True)
+
+
+def execute(command: str, cwd=None):
+    """Executes a command and returns the stdout from it"""
+    result = subprocess.run(command.split(), stdout=subprocess.PIPE, check=True, cwd=cwd)
```
chris-verclytte, line 46: What about stderr ? I would pipe it also, isn't it ?

nestor_api/lib/io.py
```diff
@@ -0,0 +1,102 @@
...
+    """Creates a copy of a directory in a temporary directory and returns its path"""
+    copy_path = get_temporary_directory_path(target_directory_prefix)
+    copy(directory_path, copy_path)
+
+    return copy_path
+
+
+def get_temporary_directory_path(prefix=""):
+    """Returns a temporary directory path with an optional prefix"""
+    random_num = math.floor(random() * 1e9)
```
chris-verclytte, line 70: Why not using native [tempfile.TemporaryDirectory](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory) ?

tests/lib/test_io.py
```diff
@@ -0,0 +1,203 @@
+# pylint: disable=missing-function-docstring disable=missing-module-docstring
```
chris-verclytte, line 1: I think missing-function-docstring and missing-module-docstring should not be disabled but fixed

chris-verclytte: Or the rules should be updated

tests/lib/test_io.py
maximejimenez: I think since we test already the `get_working_path` that retrieves the config underneath, it's not worth re-testing it again through it.

tests/lib/test_io.py
maximejimenez: Yes but the issue is that it seems that with pylint, there is no way to re-declare a partial .pylintrc for a dedicated folder such as with eslint.

nestor_api/lib/io.py
```diff
@@ -0,0 +1,102 @@
...
+
+def copy(source, destination):
+    """Copy file or directory with its content from source to destination"""
+    try:
+        shutil.copytree(source, destination)
+    except OSError as err:
+        if err.errno == errno.ENOTDIR:
+            shutil.copy(source, destination)
+        else:
+            raise
```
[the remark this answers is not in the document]
maximejimenez: https://docs.python.org/3/reference/simple_stmts.html#the-raise-statement

> If no expressions are present, raise re-raises the last exception that was active in the current scope.

nestor_api/config/config.py
```diff
@@ -0,0 +1,9 @@
+"""Nestor-api configuration"""
```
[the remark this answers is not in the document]
maximejimenez: It is the module doc-string.
We could disable it if we don't want this anymore but maybe let's do it apart from this PR.

nestor_api/lib/io.py
maximejimenez: The thing is that if this returns a not zero exit code, this instruction will throw (check=True). So I am not sure of the use case where we would need to pipe stderr.

nestor_api/lib/io.py
maximejimenez: Indeed, I did not know, but when I tried out, the issue is that it states the following :

> On completion of the context or **destruction of the temporary directory object** the newly created temporary directory and all its contents are **removed** from the filesystem.

So when I tried, since we only return the path, the object is destroyed and the directory with it 😞

chris-verclytte: Hmmmm okay, no way to make it persistent until we explicitly delete it ?

maximejimenez: In the doc it doesn't look like, it looks a bit like a smart pointer from cpp.
When no more references to the created object, it is deleted.

PR #11, opened 2020-06-25
TECH Add codeowners

chris-verclytte:


CODEOWNERS
```diff
@@ -0,0 +1 @@
+@nestor-maintainers
```
benard-g, line 1: How about putting the file under the `.github` directory ?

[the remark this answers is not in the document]
chris-verclytte: You're right, my bad

chris-verclytte: I prefer to keep it at the root for making it clear. Same for README and CONTRIBUTE

PR #12, opened 2020-06-26
Add git library

maximejimenez:
Add the git library.

I created an empty shell for the `config` library and the logger to be able to stub dependency.

I also took the time to refactor the project's config to mock it more easily.

**Note :** You can review commit by commit.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -233,7 +233,7 @@ contextmanager-decorators=contextlib.contextmanager
 # List of members which are set dynamically and missed by pylint inference
 # system, and so shouldn't trigger E1101 when accessed. Python regular
 # expressions are accepted.
-generated-members=assert_called_\w+
+generated-members=(assert_called_\w+|assert_has_\w+|call_count|mock_calls)
 
 # Tells whether missing members accessed in mixin class should be ignored. A
 # mixin class is detected if its name ends with "mixin" (case insensitive).
--- a/nestor_api/lib/git.py
+++ b/nestor_api/lib/git.py
@@ -1,0 +1,119 @@
+"""git library"""
+
+import semver
+
+import nestor_api.lib.config as config
+import nestor_api.lib.io as io
+from nestor_api.utils.logger import Logger
+
+
+def branch(repository_dir, branch_name):
+    """Checkout a branch of a repository"""
+    Logger.debug({"path": repository_dir}, "[git#branch] Repository path")
+
+    stdout = io.execute(f"git branch --list {branch_name}", repository_dir)
+
+    exists = len(stdout) > 0
+
+    io.execute(f'git checkout{"" if exists else " -b"} {branch_name}', repository_dir)
+
+
+def create_working_repository(app_name):
+    """Checkout a branch of a repository"""
+    # First, update the pristine repository for this app
+    pristine_directory = update_pristine_repository(app_name)
+
+    repository_dir = io.get_temporary_copy(pristine_directory, app_name)
+
+    Logger.debug(
+        {app_name, repository_dir},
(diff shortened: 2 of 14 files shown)


tests/config/test_config.py
```diff
@@ -0,0 +1,28 @@
+# pylint: disable=missing-function-docstring disable=missing-module-docstring
+
+from nestor_api.config.config import Configuration
+
+
+def test_get_config_path(monkeypatch):
+    monkeypatch.setenv("NESTOR_CONFIG_PATH", "/a-custom-config-path")
+    assert Configuration.get_config_path() == "/a-custom-config-path"
```
benard-g, line 8: No `test_get_config_path_configured` ?

tests/lib/test_config.py
```diff
@@ -0,0 +1,8 @@
+# pylint: disable=missing-function-docstring disable=missing-module-docstring
+
+import nestor_api.lib.config as config
+
+
+def test_for_app():
+    configuration = config.for_app("my-app")
+    assert configuration == dict({"app_name": "my-app"})
```
benard-g, line 8: I think `dict` is redundant here 🙂  (but since it is a placeholder you don't have to fix it).
```suggestion
def test_for_app():
    configuration = config.for_app("my-app")
    assert configuration == {"app_name": "my-app"}
```

tests/config/test_config.py
maximejimenez: It always needs to be configured, there is no default for this one, so there is only one test 🙂

.pylintrc
```diff
@@ -233,7 +233,7 @@ contextmanager-decorators=contextlib.contextmanager
 # List of members which are set dynamically and missed by pylint inference
 # system, and so shouldn't trigger E1101 when accessed. Python regular
 # expressions are accepted.
-generated-members=assert_called_\w+
+generated-members=(assert_called_\w+|assert_has_\w+|call_count|mock_calls)
```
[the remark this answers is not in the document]
maximejimenez: Yes, but I did not find anything better than this

nestor_api/lib/git.py
```diff
@@ -0,0 +1,119 @@
...
+
+def branch(repository_dir, branch_name):
+    """Checkout a branch of a repository"""
+    Logger.debug({"path": repository_dir}, "[git#branch] Repository path")
+
+    stdout = io.execute(f"echo $(git branch -a | grep {branch_name})", repository_dir)
+
+    exists = len(stdout) > 0
+
+    io.execute(f'git checkout{"" if exists else " -b"} {branch_name}', repository_dir)
```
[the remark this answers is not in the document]
maximejimenez: Not sure we need to force here for now, if we encounter issues on this we might change this. But for now, I would rather avoid adding by default force option everywhere.

nestor_api/lib/git.py
```diff
@@ -0,0 +1,119 @@
...
+import nestor_api.lib.config as config
+import nestor_api.lib.io as io
+from nestor_api.utils.logger import Logger
+
+
+def branch(repository_dir, branch_name):
+    """Checkout a branch of a repository"""
+    Logger.debug({"path": repository_dir}, "[git#branch] Repository path")
+
+    stdout = io.execute(f"echo $(git branch --list {branch_name})", repository_dir)
```
benard-g, line 14: Do you need the `echo` ?
```suggestion
    stdout = io.execute(f"git branch --list {branch_name}", repository_dir)
```

maximejimenez: Indeed, it is not useless, I removed it 👍

PR #13, opened 2020-06-29
Add docker library

maximejimenez:
Add the docker library to interact with docker (retrieve images, build, push).

### DIFF
--- a/nestor_api/lib/docker.py
+++ b/nestor_api/lib/docker.py
@@ -1,0 +1,82 @@
+"""Docker library"""
+
+import nestor_api.lib.config as config
+import nestor_api.lib.git as git
+import nestor_api.lib.io as io
+from nestor_api.utils.logger import Logger
+
+
+def build(app_name, context):
+    """Build the docker image of the last version of the app
+    Example:
+        > app_name = "my-app"
+        > context = {"repository": "/path_to/my-app-doc", "commit_hash": "a2b3c4"}
+        > build(app_name, context)
+    """
+    repository = context["repository"]
+    image_tag = git.get_last_tag(repository)
+
+    Logger.info(
+        {"tag": f"{app_name}@{image_tag}"}, "Application name and version",
+    )
+
+    if has_docker_image(app_name, image_tag):
+        Logger.info({}, "Docker image already built (skipped)")
+        return image_tag
+
+    app_config = config.get_app_config(app_name)
+    commit_hash = context["commit_hash"]
+    build_variables = app_config["build"]["variables"]
+
+    # Application build environment variables:
+    builds_args = []
+    for key, value in build_variables.items():
+        builds_args.append(f"--build-arg {key}={value}")
+
+    builds_args.append(f"--build-arg COMMIT_HASH={commit_hash}")
+    builds_args_str = " ".join(builds_args)
+
+    command = f"docker build --tag {app_name}:{image_tag} {builds_args_str} {repository}"
+
+    Logger.debug({"command": command}, "Docker build command")
+
+    try:
+        io.execute(command)
+    except Exception as err:
+        Logger.error({"err": err}, "Error while building Docker image")
+        raise err
+
+    return image_tag
+
+
+def has_docker_image(app_name, tag):
+    """Checks if the docker image already exists for a given app and tag"""
+    stdout = io.execute(f"docker images {app_name}:{tag} --quiet")
+    return len(stdout) != 0
+
+
+def get_registry_image_tag(app_name, image_tag, registry):
+    """Returns the image name for a given organization, app and tag"""
+    return f"{registry['organization']}/{app_name}:{image_tag}"
+
+
+def push(app_name, repository):
+    """Push an image to the configured docker registry"""
+
+    # This will need to be done a bit differently to work with GCP registry
+
+    app_config = config.get_app_config(app_name)
+
+    image_tag = git.get_last_tag(repository)
+
+    if not has_docker_image(app_name, image_tag):
+        raise RuntimeError("Docker image not available")
+
+    registry = app_config["docker"]["registry"]
+
+    # Create the tag
+    image = get_registry_image_tag(app_name, image_tag, registry)
+
+    io.execute(f"docker tag {app_name}:{image_tag} {image}")
+
+    io.execute(f"docker push {image}")
--- a/tests/lib/test_docker.py
+++ b/tests/lib/test_docker.py
@@ -1,0 +1,148 @@
+# pylint: disable=missing-function-docstring disable=missing-module-docstring
+
+import subprocess
+from unittest.mock import call
+
+import pytest
+
+import nestor_api.lib.config as config
+import nestor_api.lib.docker as docker
+import nestor_api.lib.git as git
+import nestor_api.lib.io as io
+from nestor_api.utils.logger import Logger
+
+
+def test_build_already_built(mocker):
+    mocker.patch.object(docker, "has_docker_image", return_value=True)
+    mocker.patch.object(git, "get_last_tag", return_value="1.0.0-sha-a2b3c4")
+    context = {"commit_hash": "a2b3c4", "repository": "/path_to/a_git_repository"}
+
+    image_tag = docker.build("my-app", context)
+
+    git.get_last_tag.assert_called_once_with("/path_to/a_git_repository")
+    docker.has_docker_image.assert_called_once_with("my-app", "1.0.0-sha-a2b3c4")
+
+    assert image_tag == "1.0.0-sha-a2b3c4"
+
+
+def test_build(mocker):
+    app_config = {"build": {"variables": {"var1": "val1", "var2": "val2"}}}
+
+    mocker.patch.object(docker, "has_docker_image", return_value=False)
+    mocker.patch.object(git, "get_last_tag", return_value="1.0.0-sha-a2b3c4")
+    mocker.patch.object(config, "get_app_config", return_value=app_config)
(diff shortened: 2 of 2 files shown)


nestor_api/lib/docker.py
```diff
@@ -0,0 +1,77 @@
...
+        {"tag": f"${app_name}@${image_tag}"}, "Application name and version",
+    )
+
+    if has_docker_image(app_name, image_tag):
+        Logger.info({}, "Docker image already built (skipped)")
+        return image_tag
+
+    app_config = config.get_app_config(app_name)
+    commit_hash = context.get("commit_hash")
+    build_variables = app_config.get("build").get("variables")
```
benard-g, line 24: I would recommend to use `app_config["build"]["variables"]` to have a more explicit error if the keys are missing.
Same thing for all `dict` access on required keys.

nestor_api/lib/docker.py
```diff
@@ -0,0 +1,77 @@
...
+        io.execute(command)
+    except Exception as err:
+        Logger.error({"err": err}, "Error while building Docker image")
+        raise err
+
+    return image_tag
+
+
+def has_docker_image(app_name, tag):
+    """Checks if the docker images already exists for a given app and tag"""
```
benard-g, line 48:
```suggestion
    """Checks if the docker image already exists for a given app and tag"""
```

nestor_api/lib/docker.py
```diff
@@ -0,0 +1,77 @@
...
+    return len(stdout) != 0
+
+
+def get_registry_image_tag(app_name, image_tag, registry):
+    """Returns the image name for a given organization, app and tag"""
+    return f"{registry.get('organization')}/{app_name}:{image_tag}"
+
+
+def push(app_name, repository):
+    """Push the image an image to the configured docker registry"""
```
benard-g, line 59:
```suggestion
    """Push an image to the configured docker registry"""
```

nestor_api/lib/docker.py
```diff
@@ -0,0 +1,77 @@
...
+    builds_args = []
+    for key, value in build_variables.items():
+        builds_args.append(f"--build-arg {key}={value}")
+
+    builds_args.append(f"--build-arg COMMIT_HASH={commit_hash}")
+    builds_args_str = ' '.join(builds_args)
+
+    command = f'docker build --tag {app_name}:{image_tag} {builds_args_str} {repository}'
+
+    Logger.debug({command}, "Docker build command")
```
benard-g, line 36: Be careful, this creates a `set` and not a `dict` (js again 🙂 )
```suggestion
    Logger.debug({'command': command}, "Docker build command")
```

tests/lib/test_docker.py
```diff
@@ -0,0 +1,151 @@
...
+import nestor_api.lib.docker as docker
+import nestor_api.lib.git as git
+import nestor_api.lib.io as io
+from nestor_api.utils.logger import Logger
+
+
+def test_build_already_built(mocker):
+    mocker.patch.object(docker, "has_docker_image", return_value=True)
+    mocker.patch.object(git, "get_last_tag", return_value="1.0.0-sha-a2b3c4")
+    context = {"commitHash": "a2b3c4", "repository": "/path_to/a_git_repository"}
```
benard-g, line 18:
```suggestion
    context = {"commit_hash": "a2b3c4", "repository": "/path_to/a_git_repository"}
```

tests/lib/test_docker.py
```diff
@@ -0,0 +1,151 @@
...
+
+def test_build(mocker):
+    app_config = {"build": {"variables": {"var1": "val1", "var2": "val2"}}}
+
+    mocker.patch.object(docker, "has_docker_image", return_value=False)
+    mocker.patch.object(git, "get_last_tag", return_value="1.0.0-sha-a2b3c4")
+    mocker.patch.object(config, "get_app_config", return_value=app_config)
+    mocker.patch.object(io, "execute", return_value="")
+
+    context = {"commitHash": "a2b3c4", "repository": "/path_to/a_git_repository"}
```
benard-g, line 36:
```suggestion
    context = {"commit_hash": "a2b3c4", "repository": "/path_to/a_git_repository"}
```

tests/lib/test_docker.py
```diff
@@ -0,0 +1,151 @@
...
+
+    git.get_last_tag.assert_called_once_with("/path_to/a_git_repository")
+    docker.has_docker_image.assert_called_once_with("my-app", "1.0.0-sha-a2b3c4")
+    config.get_app_config.assert_called_once_with("my-app")
+    io.execute.assert_called_once_with(
+        "docker build"
+        " --tag my-app:1.0.0-sha-a2b3c4"
+        " --build-arg var1=val1"
+        " --build-arg var2=val2"
+        " --build-arg COMMIT_HASH=None"
```
benard-g, line 48: `None` ?

tests/lib/test_docker.py
```diff
@@ -0,0 +1,151 @@
...
+def test_build_failure(mocker):
+    app_config = {"build": {"variables": {"var1": "val1", "var2": "val2"}}}
+
+    mocker.patch.object(docker, "has_docker_image", return_value=False)
+    mocker.patch.object(git, "get_last_tag", return_value="1.0.0-sha-a2b3c4")
+    mocker.patch.object(config, "get_app_config", return_value=app_config)
+
+    exception = subprocess.CalledProcessError(1, "docker build")
+    mocker.patch.object(io, "execute", side_effect=[exception])
+    mocker.patch.object(Logger, "error", return_value="")
```
benard-g, line 64:
```suggestion
    mocker.patch.object(Logger, "error")
```

tests/lib/test_docker.py
```diff
@@ -0,0 +1,151 @@
...
+
+    context = {"commitHash": "a2b3c4", "repository": "/path_to/a_git_repository"}
```
benard-g, line 66:
```suggestion
    context = {"commit_hash": "a2b3c4", "repository": "/path_to/a_git_repository"}
```

tests/lib/test_docker.py
```diff
@@ -0,0 +1,151 @@
...
+
+    git.get_last_tag.assert_called_once_with("/path_to/a_git_repository")
+    docker.has_docker_image.assert_called_once_with("my-app", "1.0.0-sha-a2b3c4")
+    config.get_app_config.assert_called_once_with("my-app")
+    io.execute.assert_called_once_with(
+        "docker build"
+        " --tag my-app:1.0.0-sha-a2b3c4"
+        " --build-arg var1=val1"
+        " --build-arg var2=val2"
+        " --build-arg COMMIT_HASH=None"
```
benard-g, line 79: `None` again ?

tests/lib/test_docker.py
```diff
@@ -0,0 +1,151 @@
...
+    mocker.patch.object(io, "execute", return_value="")
+
+    with pytest.raises(RuntimeError):
+        docker.push("my-app", "/path_to/a_git_repository")
+
+    config.get_app_config.assert_called_once_with("my-app")
+    git.get_last_tag.assert_called_once_with("/path_to/a_git_repository")
+    io.execute.assert_has_calls([
+        call('docker images my-app:1.0.0-sha-a2b3c4 --quiet'),
+    ])
```
benard-g, line 132: Why not `assert_called_once_with` ?

tests/lib/test_docker.py
```diff
@@ -0,0 +1,151 @@
...
+    io.execute.assert_called_once_with(
+        "docker build"
+        " --tag my-app:1.0.0-sha-a2b3c4"
+        " --build-arg var1=val1"
+        " --build-arg var2=val2"
+        " --build-arg COMMIT_HASH=None"
+        " /path_to/a_git_repository"
+
+    )
+    Logger.error.assert_called_once_with(
```
Antoninbln, line 83: I'm wondering if this is possible to assert the error raised from `docker.build()` instead of testing that `logger.error()` is called ? What do you think ?

maximejimenez: It is what is done here : https://github.com/ChauffeurPrive/nestor-api/blob/docker-lib/tests/lib/test_docker.py#L68

nestor_api/lib/docker.py
```diff
@@ -0,0 +1,82 @@
...
+import nestor_api.lib.config as config
+import nestor_api.lib.git as git
+import nestor_api.lib.io as io
+from nestor_api.utils.logger import Logger
+
+
+def build(app_name, context):
+    """Build the docker image of the last version of the app
+        Example:
+        >>> app_name = {"my-app"}
```
benard-g: I thought `app_name` would be a string. Is a set what you a expect as an argument ?

tests/lib/test_docker.py
```diff
@@ -0,0 +1,148 @@
...
+
+def test_push_no_image(mocker):
+    mocker.patch.object(
+        config, "get_app_config",
+        return_value={"docker": {"registry": {"organization": "my-organization"}}}
+    )
+    mocker.patch.object(git, "get_last_tag", return_value="1.0.0-sha-a2b3c4")
+    mocker.patch.object(io, "execute", return_value="")
+
+    with pytest.raises(RuntimeError):
```
Kenny407, line 124: Question regarding the test of the exceptions, given that `RuntimeError` is a bit general, is there a way to assert what the message of the exception is?

Disclaimer: I haven't used python that much so I'm just wondering 😅

maximejimenez: Yes, indeed, I added this assertion to ensure the right error is raised 👍

PR #14, opened 2020-06-29
Add config library

benard-g:
* Add coverage summary in the console
* Add the `config` library

You may review commit by commit 😉   (the 2nd one is only code auto-formatting)


tests/lib/test_config.py
```diff
@@ -1,8 +1,87 @@
...
+    )
+
+
+@pytest.mark.usefixtures("config")
+def test_create_temporary_directory(mocker):
+    spy = mocker.patch.object(io, "create_temporary_copy")
+
+    config.create_temporary_config_copy()
+
+    spy.assert_called_once_with("/fixtures-nestor-config", "config")
```
maximejimenez, line 34: Maybe assert on the return value because it is almost as important as the rest, otherwise you cannot guess where the temporary copy has been created 🙂

tests/lib/test_config.py
```diff
@@ -1,8 +1,87 @@
...
+from nestor_api.lib.errors import AppConfigurationNotFoundError
+import nestor_api.lib.io as io
+
+
+def test_change_environment(mocker):
+    spy = mocker.patch.object(io, "execute")
+
+    config.change_environment("environment", "path/to/config")
+
+    spy.assert_has_calls(
```
Kenny407, line 18: I'm wondering if you should assert on the return values of the tests on this file, seems like we are only spying the calls. Just for consistency maybe you can add some asserts to the return values?

nestor_api/lib/config.py
```diff
@@ -1,9 +1,60 @@
...
+    return io.create_temporary_copy(Configuration.get_config_path(), "config")
+
+
+def get_app_config(app_name: str) -> dict:
+    """Load the configuration of an app"""
+    app_config_path = os.path.join(Configuration.get_config_path(), APPS_FOLDER, f"{app_name}.yaml")
+    if not io.exists(app_config_path):
+        raise AppConfigurationNotFoundError(app_name)
+
+    app_config = io.from_yaml(app_config_path)
```
Kenny407, line 33: I think we will need to change this eventually because from PyYAML, the parser that uses `safe_load` silently manages bad yaml (like duplicates), I was exploring another library like https://pypi.org/project/ruamel.yaml/ to handle those cases in a more specific manner. Or to override and add constructor to handle cases that PyYAML is not handling properly

nestor_api/lib/config.py
```diff
@@ -1,9 +1,60 @@
...
+    environment_config = get_environment_config()
+
+    config = dict_utils.deep_merge(environment_config, app_config)
+
+    # To be done: Validate configuration using nestor-config-validator
+
+    return __map_values(config)
+
+
+def get_environment_config() -> dict:
```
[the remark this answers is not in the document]
benard-g: I will go back to `project` then

tests/lib/test_config.py
```diff
@@ -1,8 +1,88 @@
...
+            call("git stash", "path/to/config"),
+            call("git fetch origin", "path/to/config"),
+            call("git checkout environment", "path/to/config"),
+            call("git reset --hard origin/environment", "path/to/config"),
+        ]
+    )
+
+
+@pytest.mark.usefixtures("config")
+def test_create_temporary_directory(mocker):
```
maximejimenez, line 29:
```suggestion
def test_create_temporary_config_copy(mocker):
```

PR #15, opened 2020-07-01
Add variable reference resolution in lib-config

benard-g:
Allow the usage of value references with the pattern `{{...}}` in the configuration.

### Example

The config :
```yaml
domain: 'website.com'
app_uri: 'app.{{domain}}'
```
will be resolved as
```yaml
domain: 'website.com'
app_uri: 'app.website.com'
```


nestor_api/utils/dict.py
```diff
@@ -15,8 +15,7 @@ def deep_merge(destination: dict, source: dict) -> dict:
 
     def _deep_merge_rec(dest, src):
         for key in src:
-            # pylint: disable=bad-continuation
-            if key in dest and isinstance(dest.get(key), dict) and isinstance(src[key], dict):
+            if key in dest and isinstance(dest[key], dict) and isinstance(src[key], dict):
```
Olivier-Daire, line 18:
```suggestion
            if isinstance(dest.get(key), dict) and isinstance(src[key], dict):
```
If key is not in dict, it will return `None`, no need for the extra check 😉

nestor_api/utils/dict.py
```diff
@@ -15,8 +15,7 @@ def deep_merge(destination: dict, source: dict) -> dict:
 
     def _deep_merge_rec(dest, src):
         for key in src:
-            # pylint: disable=bad-continuation
-            if key in dest and isinstance(dest.get(key), dict) and isinstance(src[key], dict):
+            if isinstance(dest.get(key), dict) and isinstance(src[key], dict):
```
maximejimenez, line 18: Why do we access 2 dicts with 2 different ways : `[]` & `get`

benard-g: Since we are iterating over `src`, the key can be missing in `dest` and in this case the `[]` operator will throw whereas the `.get()` will return `None`.
In this case I removed the first check which was redundant.

nestor_api/lib/config.py
```diff
@@ -66,7 +67,7 @@ def _resolve_variable(template: str, variables: dict) -> str:
         var_value = variables.get(var_name)
         if var_value is not None:
             if not isinstance(var_value, str):
-                raise InvalidConfigurationKeyReference(var_name)
+                raise ConfigurationError(path, "variable should be a string")
```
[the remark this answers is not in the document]
benard-g: I will go with `"Referenced variable should resolve to a string"`.

PR #16, opened 2020-07-02
Add config validator lib

Kenny407:
### Purpose of this PR
This PR adds a library for validating configuration for applications and projects. By using [jsonschema](https://python-jsonschema.readthedocs.io/en/stable/).

### How to review?
Commit by commit supported

### Remarks
If we use `yaml` as the library to parse yaml files, I have noticed that we can't detect by default if there's a bad misconfiguration such as: having duplicated keys, for instance, or others

For example, this is an invalid yaml.
```yaml
app: 'test'
app: 'random'
```

The current library we are using is silently taking `app: 'random'` as the value and not raising any exception.

I've read a bit and we could override some constructors to raise our own exceptions and detect this type of issues. Another alternative is to use [ruamel.yaml](https://pypi.org/project/ruamel.yaml/) which seems to comply with the yaml spec.


.pylintrc
```diff
@@ -61,6 +61,9 @@ confidence=
 # no Warning level messages displayed, use "--disable=all --enable=classes
 # --disable=W".
 disable=print-statement,
+        line-too-long,
+        bad-continuation,
+        missing-function-docstring,
```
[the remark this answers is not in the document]
Kenny407: From what I understand `line-too-long` and `bad-continuation` will continue to be checked in the project. Given that this is under the `[MESSAGES CONTROL]` [messaging control in pylint](http://pylint.pycqa.org/en/latest/user_guide/message-control.html) seems like then flag the properties that sometimes we want to enable/disable

Example, if I don't add the lines here in this section
```py
long_line = "abcdfgh" # pylint disable=line-too-long
```
Pylint would still lint the `line-too-long` rule.

validator/validate.py
```diff
@@ -0,0 +1,78 @@
...
+
+def build_app_path():
+    """Builds the path for /apps folder"""
+    target_path = Configuration.get_target_path() or os.path.dirname(__file__)
+    return os.path.join(target_path, "apps")
+
+
+def build_project_conf_path():
+    """Builds the path for project.yaml"""
+    target_path = Configuration.get_target_path() or os.path.dirname(__file__)
```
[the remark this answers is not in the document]
Kenny407: I'm not exactly sure about giving a default value to this one. Perhaps if it's not set then I can raise a custom error?

.pylintrc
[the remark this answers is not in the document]
Olivier-Daire: This will disable the rules for the entire project, without the need of a `pylint: disable`.

The comment `pylint: disable` itself should work without anything else, for example we use [`# pylint: disable=broad-except`](https://github.com/ChauffeurPrive/nestor-api/blob/09284992c5552a007d84bca95b6010a0bca69c8c/nestor_api/lib/git.py#L101) which is not in the `.pylintrc` file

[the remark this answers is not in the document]
Olivier-Daire: I don't really have a better approach, maybe use Pycharm Code > reformat code tool, it usually handles properly long lines and indenting

validator/validate.py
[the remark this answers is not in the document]
Olivier-Daire: Yes, I think that is an acceptable solution

.pylintrc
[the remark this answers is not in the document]
Kenny407: Solved 🙏  thanks !

validator/schemas/application.py
```diff
@@ -0,0 +1,199 @@
...
+                "tplTeam": {
+                    "type": "string",
+                    "example": "team-payment",
+                    "description": "Define the team that is owner.",
+                },
+                "tplSlackAlertingChannel": {
+                    "type": "string",
+                    "example": "team-payment-alerts",
+                    "description": "Define the slack channel for alerting",
+                },
```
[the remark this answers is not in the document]
Kenny407: I'll keep them but will make `slack` as optional 👍

validator/validate.py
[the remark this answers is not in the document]
Kenny407: Done ✅

Kenny407:
@Olivier-Daire I addressed all your comments 🙏  let me know what you think! thanks for reviewing it

validator/schemas/application.py
```diff
@@ -0,0 +1,199 @@
+# pylint: disable=duplicate-code
+"""The application schema configuration managed by Nestor"""
+
+APPLICATION_SCHEMA = {
+    "type": "object",
+    "properties": {
+        "app": {"type": "string", "maxLength": 58,},
```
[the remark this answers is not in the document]
Kenny407: Currently it's only used on this schema, and the idea is the others would inherit from this one (if any), shall I still move it?

validator/validate.py
```diff
@@ -0,0 +1,80 @@
...
+        raise InvalidTargetPathError(target_path)
+    return os.path.join(target_path, "apps")
+
+
+def build_project_conf_path():
+    """Builds the path for project.yaml"""
+    target_path = Configuration.get_target_path()
+    if target_path is None:
+        raise InvalidTargetPathError(target_path)
+    return os.path.join(target_path, "project.yaml")
```
maximejimenez, line 27: I think there is some similar function in `nestor-api/lib/config`

validator/validate.py
```diff
@@ -0,0 +1,80 @@
...
+    Returns:
+        String: The validated file
+    """
+    with open(file_path, "r") as file:
+        yaml_file_data = file.read()
+        yaml_file = yaml.safe_load(yaml_file_data)
+        return jsonschema.validate(yaml_file, schema)
+
+
+def validate():
```
maximejimenez, line 46: I would change the prototype to have passed the validation target and the file to validate to have something like
```suggestion
def validate(file_to_validate, target_validation):
```

This way you will be able to call this in the config lib 🙂 
So it means rewriting a bit this function but it will way easier to call afterwards.

validator/errors/errors.py
```diff
@@ -0,0 +1,22 @@
+"""Custom Errors for Nestor Config Validator"""
```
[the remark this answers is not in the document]
Olivier-Daire: We talked about possibly having the validator as a standalone module, in this case it could be interesting to keep errors inside `validator`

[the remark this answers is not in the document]
Kenny407: If we have the validator as a standalone module as Olivier said, then it would make sense to have it's own errors and also make it part of the commands for `Makefile`, etc.

validator/validate.py
Kenny407: If we plan to do it then standalone then we can keep these ones, no?

validator/schemas/application.py
```diff
@@ -0,0 +1,181 @@
...
+                "tplExpandedTimeout": {
+                    "type": "integer",
+                    "example": "3600",
+                    "description": "Define the timeout on the ingress controller",
+                },
+                "tplMountDirectory": {
+                    "type": "string",
+                    "example": "/opt/my-folder",
+                    "description": "Define the folder to mount a file through ConfigMap or others",
+                },
```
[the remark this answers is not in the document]
Kenny407: Maybe it's just a new path that kubernetes would read? not sure about the usage of this one, maybe we can get rid of it

.pylintrc
```diff
@@ -357,7 +357,7 @@ ignore-docstrings=yes
 ignore-imports=no
 
 # Minimum lines number of a similarity.
-min-similarity-lines=4
+min-similarity-lines=10
```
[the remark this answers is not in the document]
Kenny407: It was needed due to the similarity of code between the schemas but with the new updates I will revert the change.

validator/validate.py
Kenny407: There's already a `validate_file` function in this file, just above that one. One validates file and this one validates paths, I will rename it to make it more specific 👍

tests/config_validator/test_validate.py
```diff
@@ -0,0 +1,85 @@
...
+from jsonschema import ValidationError
+import pytest
+
+from tests.__fixtures__.example_schema import EXAMPLE_SCHEMA
+from validator.config.config import Configuration
+from validator.errors.errors import InvalidTargetPathError
+import validator.validate as config_validator
+
+
+def test_build_app_path_with_env(monkeypatch):
```
benard-g, line 15: Is it required to have the argument sometimes named `monkeypatch` and sometimes `mocker` ?
I would personally find it clearer if we always used `mocker` but I don't know if this is a common practice.

validator/config/config.py
```diff
@@ -0,0 +1,32 @@
...
+import os
+
+
+class Configuration:
+    """Nestor_config_validator configuration"""
+
+    @staticmethod
+    def get_target_path():
+        """Returns the target path of the project configuration"""
+        return os.getenv("NESTOR_TARGET_PATH")
```
benard-g: It could be good to use `NESTOR_CONFIG_PATH` here has it probably serves the same purpose 🙂  (indicating where the directory holding the configuration is).

validator/errors/errors.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class InvalidTargetPathError(Exception):
+    """Exception raised when the target path (NESTOR_TARGET_PATH) env is not set"""
```
benard-g: Just a thought but what about having the exception raised directly from `get_target_path()` ?
It would solve the duplicate you have with `build_app_path`/`build_project_conf_path` being too similar.

I don't know if there are cases where the variable being missing can be acceptable 🤷

validator/errors/errors.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def __init__(self, target_path: str):
+        super().__init__(
+            f"The environment variable NESTOR_TARGET_PATH is not set. Current value: ${target_path}"
+        )
+
+
+class ConfigurationNotValidError(Exception):
```
benard-g, line 13: Is this error class used ? (or will it be used later ?)

validator/errors/errors.py
```diff
@@ -0,0 +1,22 @@
...
+    """Exception raised when a configuration is invalid"""
+
+    def __init__(self, validation_errors):
+        self.validation_errors = validation_errors
+        self.message = "Please fix the configuration with the help of the validationErrors property"
+        super().__init__(self.validation_errors)
```
benard-g, line 19: I would have expected `super` to be called with a message. What is the type of `validation_errors` ?

validator/schemas/project.py
```diff
@@ -0,0 +1,88 @@
...
+                                },
+                            },
+                        },
+                    },
+                },
+            },
+            "required": ["build", "registries"],
+            "additionalProperties": False,
+        },
+        # TODO to be defined pylint: disable=fixme
```
benard-g, line 44: FYI we used "Awaiting for implementation" to go around this warning in [some other files](https://github.com/ChauffeurPrive/nestor-api/blob/master/nestor_api/lib/app.py#L7) 😅

validator/validate.py
```diff
@@ -0,0 +1,83 @@
...
+
+import jsonschema
+import yaml
+
+from validator.config.config import Configuration, SupportedValidations
+from validator.errors.errors import InvalidTargetPathError
+from validator.schemas.schema import SCHEMAS
+
+
+def build_app_path():
```
benard-g, line 14:
```suggestion
def build_apps_path():
```
```suggestion
def build_apps_root_path():
```
Or the same with `get_` 🤷  ?

validator/validate.py
```diff
@@ -0,0 +1,83 @@
...
+    Returns:
+        String: The validated file
+    """
+    with open(file_path, "r") as file:
+        yaml_file_data = file.read()
+        yaml_file = yaml.safe_load(yaml_file_data)
+        return jsonschema.validate(yaml_file, schema)
+
+
+def validate_deployment_files():
```
benard-g, line 46: I don't know the use case but why not receiving the `NESTOR_VALIDATION_TARGET` as an argument here ?

tests/config_validator/test_validate.py
```diff
@@ -0,0 +1,85 @@
...
+    yaml_fixture_path = Path(
+        os.path.dirname(__file__), "..", "__fixtures__", "example_valid_config.yaml"
+    ).resolve()
+
+    # Validate an exception is not raised on valid schemas
+    try:
+        result = config_validator.validate_file(yaml_fixture_path, EXAMPLE_SCHEMA)
+        assert result is None
+    except ValidationError as error:
+        pytest.fail(f"It should have not raised an exception on a valid file ${error}")
```
benard-g, line 49: Is the try/except required here ? (I imagine that if an exception is raised the test will fail)

validator/config/config.py
```diff
@@ -0,0 +1,32 @@
...
+    def get_validation_target():
+        """Returns the type of validation target"""
+        return os.getenv("NESTOR_VALIDATION_TARGET")
+
+
+class SupportedValidations(Enum):
+    """Enum of Supported validations by Nestor
+
+    Args:
+        Enum (Enum): Generic enumeration class
```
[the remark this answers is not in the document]
Kenny407: Yes it was required I didn't want to disable it just for the argument. It doesn't hurt that much

validator/config/config.py
```diff
@@ -0,0 +1,32 @@
...
+
+    @staticmethod
+    def get_target_path():
+        """Returns the target path of the project configuration"""
+        return os.getenv("NESTOR_TARGET_PATH")
+
+    @staticmethod
+    def get_validation_target():
+        """Returns the type of validation target"""
+        return os.getenv("NESTOR_VALIDATION_TARGET")
```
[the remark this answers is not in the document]
Kenny407: Will add a new readme, thanks for the catch.

tests/config_validator/test_validate.py
Kenny407: The concepts are similar but the monkeypatching is new to me, here's a link that helps to clarify the difference [here](https://stackoverflow.com/a/47101194/9492453). With that said, I will rename it to `mock` given I'm mocking and not monkeypatching

validator/errors/errors.py
Kenny407: It will be used later. I'm aiming to enhance the way we report errors.

validator/errors/errors.py
Kenny407: This will be an array with a specific error for each file, I will specify it in the next PR 👍

tests/config_validator/test_validate.py
Kenny407: As for now the good validations will return a `None` so this is just a way of asserting the validation passed.

tests/__fixtures__/validator/projects/project.yaml
```diff
@@ -0,0 +1,42 @@
+processes:
+  - name: web
+    start_command: npm run start
+    is_cronjob: false
```
Olivier-Daire, line 4: I might be a bit nazi but I think that having `app`, `git` and `is_enable` first makes it more straightforward to read

tests/config_validator/test_validate.py
```diff
@@ -76,10 +76,22 @@ def test_validate_error_validation_target(mocker):
...
+def test_validate_projects(mocker):
+    real_config_fixture_path = Path(
+        os.path.dirname(__file__), "..", "__fixtures__", "validator", "projects", "project.yaml"
+    ).resolve()
+    mocker.patch.object(config_validator, "build_apps_path", return_value=real_config_fixture_path)
+    mocker.patch.object(
+        config_validator, "build_project_conf_path", return_value=real_config_fixture_path
+    )
+    mocker.patch.object(Configuration, "get_validation_target", return_value="PROJECTS")
+    config_validator.validate_deployment_files()
```
Olivier-Daire, line 97: I find it a bit weird at first sight to not have an assert and I am afraid that the stacktrace might not be super helpful in case of an exception raised by `validate_deployment_files()`

What are your thoughts on doing the following?
```python
try:
    config_validator.validate_deployment_files()
except Exception as e:
    self.fail(f"validate_deployment_files() raised an unexpected Exception {err}")
```

validator/validate.py
```diff
@@ -64,12 +63,19 @@ def validate_deployment_files():
...
         # Validate each application file
+
+        # We will only validate .yaml or .yml files
         files_in_dir = [
-            f for f in os.listdir(apps_path) if os.path.isfile(f) and not f.startswith(".")
+            f.lower()
+            for f in os.listdir(apps_path)
+            if os.path.isfile(os.path.join(apps_path, f))
+            and not f.startswith(".")
+            and (f.endswith(".yaml") or f.endswith(".yml"))
```
Olivier-Daire, line 73: Maybe we could move this to a function `is_yaml_file(file)`

tests/config_validator/test_validate.py
Kenny407: Yes I agree with you, I will add it to make the test clearer. FYI, we already thought of an improvement and it's to accumulate the errors for each file and not fail on the first invalid file. And if everything is fine we will assert on a more reasonable response.

PR #17, opened 2020-07-06
Fix codeowners format

chris-verclytte:


CODEOWNERS
```diff
@@ -1 +1 @@
-*   @nestor-maintainers
+*       @nestor-maintainer
```
Kenny407, line 1: Isn't the name [nestor-maintainers](https://github.com/orgs/ChauffeurPrive/teams/nestor-maintainers) ?

benard-g: Try also @ChauffeurPrive/nestor-maintainers (that is what Github suggests me)

PR #19, opened 2020-07-06
Move git lib test to unittest framework

chris-verclytte:
Enrich Git lib with clone and get_branch_hash features and move testing to unittest.

### DIFF
--- a/nestor_api/lib/git.py
+++ b/nestor_api/lib/git.py
@@ -7,6 +7,10 @@ import nestor_api.lib.io as io
 from nestor_api.utils.logger import Logger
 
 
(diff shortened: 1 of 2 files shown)


nestor_api/lib/git.py
```diff
@@ -7,6 +7,11 @@
 from nestor_api.utils.logger import Logger
 
 
+def clone(repository: str, remote_repository_url: str, branch_name: str = "master"):
+    """Clone a repository branch locally"""
+    return io.execute(f"git clone {remote_repository_url} -b {branch_name}", repository)
```
benard-g, line 12: `update_repository` already performs the `clone` operation. Not sure if we want to differentiate it or not.

PR #29, opened 2020-07-07
Replace pytest by unittest framework in config lib

chris-verclytte:


tests/lib/test_config.py
```diff
@@ -1,176 +1,214 @@
...
-def test_get_app_config(mocker):
-    mocker.patch.object(Configuration, "get_config_path", return_value="tests/__fixtures__/config")
+        path = config.create_temporary_config_copy()
 
-    app_config = config.get_app_config("backoffice")
+        io_mock.create_temporary_copy.assert_called_once_with(
+            "/fixtures-nestor-config",
+            "config"
+        )
+        assert path == "/temporary/path"
```
Olivier-Daire, line 39: Since we are now using unittest, I think we could use the more classical syntax [`self.assertEqual()`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertEqual)

```suggestion
        self.assertEqual(path, "/temporary/path")
```

tests/lib/test_config.py
```diff
@@ -1,176 +1,206 @@
...
+        io_mock.create_temporary_copy.assert_called_once_with("/fixtures-nestor-config", "config")
+        assert path == "/temporary/path"
+
+    @patch("nestor_api.lib.config.get_project_config", autospec=True)
+    @patch("nestor_api.lib.config.Configuration", autospec=True)
+    def test_get_app_config(self, configuration_mock, get_project_config_mock, io_mock):
+        configuration_mock.get_config_path.return_value = "tests/__fixtures__/config"
+        configuration_mock.get_config_app_folder.return_value = "apps"
+        io_mock.exists.return_value = True
+        io_mock.from_yaml.return_value = {
```
Olivier-Daire, line 47: I think that this is something we should talk about and all agree on, but I find that these "mock configurations" are a bit polluting the test code. 
There is an [interesting discussion about the options to mock multiple methods ](https://stackoverflow.com/questions/29318987/preferred-way-of-patching-multiple-methods-in-python-unit-test).

Personally I think that I would be more in favor of using 
```python
@patch.multiple('nestor_api.lib.config.Configuration', 
                get_config_path=MagicMock(return_value='tests/__fixtures__/config'), 
                get_config_app_folder=MagicMock(return_value='apps'))
```

Let me know what you think

tests/lib/test_config.py
```diff
@@ -1,176 +1,206 @@
...
+            assert err.value.errors[2].message == "Referenced variable should resolved to a string"
+
+            assert err.value.errors[3].path == "CONFIG.deep_dict.sub_dict.b"
+            assert err.value.errors[3].message == "Referenced variable should resolved to a string"
+
+            assert err.value.errors[4].path == "CONFIG.deep_dict.sub_array[0].a"
+            assert err.value.errors[4].message == "Referenced variable should resolved to a string"
+
+            assert err.value.errors[5].path == "CONFIG.deep_dict.sub_array[1].b"
+            assert err.value.errors[5].message == "Referenced variable should resolved to a string"
```
Olivier-Daire, line 206: I know that it was already done this way but couldn't we test the whole array instead of testing each item?

tests/lib/test_config.py
chris-verclytte: I understand your point but I think it is just moving the problem to decorators that will also clutter the code and if you have a return value that is quite bit (a dict for example), then it would be a little bit weird to have it in the patch section :/ 
I know that fixture in pytest are definitely better approach for this point... 🙄 

Usually when the test function body become important I used to separate them with comment to make it clearer like this:
```python
def myFunction():
	# Mocks
    ...
	# Test
	...
	# Assertions
	...
```
What do you think ?

Olivier-Daire: I think that it would not clutter the code that much if big return values are fixtures (either in another file or just a variable in the test file).
But this brings another debate / problem of not having fixtures right next to the test case.

Yes, at least comments, imo, so that it is easy to skim through the test

chris-verclytte: Ok I've gone with comments but understand your point and we should discuss about it. I create a note in the board :)

PR #34, opened 2020-07-08
Replace pytest by unittest framework in io lib

chris-verclytte:


tests/lib/test_io.py
```diff
@@ -1,206 +1,166 @@
...
-def test_execute_should_raise_if_failure():
-    with pytest.raises(Exception):
+# pylint: disable=no-self-use
+class TestIoLib(TestCase):
+    @patch("nestor_api.lib.io.shutil", autospec=True)
+    def test_copy_single_file(self, shutil_mock):
+        shutil_mock.copytree.side_effect = OSError(errno.ENOTDIR, "some reason")
+        io.copy("source/path/file.yaml", "dest/path/file.yaml")
+        shutil_mock.copytree.assert_called_once()
+        shutil_mock.copy.assert_called_with("source/path/file.yaml", "dest/path/file.yaml")
```
benard-g, line 17: I personally think that it is clearer to keep the blank lines between the 3 parts of the test as you did in `test_get_working_path()` (no need for the comments tho). What do you think ?
Same comment for all other tests.
```suggestion
    def test_copy_single_file(self, shutil_mock):
        shutil_mock.copytree.side_effect = OSError(errno.ENOTDIR, "some reason")

        io.copy("source/path/file.yaml", "dest/path/file.yaml")

        shutil_mock.copytree.assert_called_once()
        shutil_mock.copy.assert_called_with("source/path/file.yaml", "dest/path/file.yaml")
```

tests/lib/test_io.py
```diff
@@ -1,206 +1,166 @@
...
+        path_mock.return_value.exists.return_value = False
+        result = io.exists("unexisting/path")
+        path_mock.assert_called_once_with("unexisting/path")
+        path_mock.return_value.exists.assert_called_once()
+        self.assertFalse(result)
+
+    @patch("nestor_api.lib.io.yaml", autospec=True)
+    def test_from_yaml(self, yaml_mock):
+        with patch("nestor_api.lib.io.open", mock_open(read_data="some yaml")) as open_mock:
+            yaml_mock.safe_load.return_value = "parsed yaml"
```
benard-g, line 85: Maybe return a `dict` here as the function is not supposed to return a string.

tests/lib/test_io.py
```diff
@@ -1,206 +1,166 @@
...
+            open_mock.assert_called_with("example.yml", "r")
+            yaml_mock.safe_load.assert_called_once_with("some yaml")
+            self.assertEqual(parsed_yaml, "parsed yaml")
+
+    @patch("nestor_api.lib.io.Configuration", autospec=True)
+    @patch("nestor_api.lib.io.os", autospec=True)
+    def test_get_pristine_path(self, os_mock, configuration_mock):
+        # Mocks
+        configuration_mock.get_pristine_path.return_value = "/tmp/nestor/pristine"
+        os_mock.path.join.return_value = "/tmp/nestor/pristine/my_path_name"
```
benard-g, line 98: I am not sure if mocking `os.path.join()` is a bit too much or not 🤔.

tests/lib/test_io.py
Kenny407: I think it was agreed with @Olivier-Daire and @chris-verclytte to keep the comments just to give it format to the tests. Also I think it will be up to discussion whether we do multiple.patch to avoid cluttering the mock part but I'm no expert here 🙉

chris-verclytte: Yes but for such a small test, I don't think it is necessary

chris-verclytte:
@Kenny407 @benard-g Took your feedback, merging it.

tests/lib/test_io.py
```diff
@@ -1,204 +1,160 @@
-import os
-from pathlib import Path
-import shutil
-import subprocess
+# pylint: disable=missing-class-docstring disable=missing-function-docstring disable=missing-module-docstring
```
benard-g, line 1: I think you can remove them since https://github.com/ChauffeurPrive/nestor-api/pull/33 got merged

tests/lib/test_io.py
```diff
@@ -1,204 +1,160 @@
...
-    io.execute("a command with --arg1 arg-value")
-
-    subprocess.run.assert_called_with(
-        ["a", "command", "with", "--arg1", "arg-value"], check=True, cwd=None, stdout=-1
-    )
-
-
-def test_execute_should_raise_if_failure():
-    with pytest.raises(Exception):
+# pylint: disable=no-self-use
```
benard-g, line 10: Same here

PR #35, opened 2020-07-08
Replace pytest by unittest framework in docker lib

chris-verclytte:


.pylintrc
```diff
@@ -138,7 +138,8 @@ disable=print-statement,
         xreadlines-attribute,
         deprecated-sys-function,
         exception-escape,
-        comprehension-escape
+        comprehension-escape,
+        bad-continuation
```
Olivier-Daire, line 142: I am not sure about adding `bad-continuation` here, most of the time it is just a matter of formatting, and it helps keeping the code more readable

chris-verclytte: Yep but it is definitely putting a false positive in our code with black behaviour and they don't agree to change the way it is working in black (see the comment description for link to both discussion on pylint and black repositories). The only workaround would be to put a local comment for disabling it. If you prefer, I can go with this way. Let me know.

Olivier-Daire: Yes, if it's an edge case if think it might be worth putting a comment and not disabling it for the whole test module

PR #39, opened 2020-07-10
Update the way we pass app_config to the lib functions

benard-g:
In this PR, I changed the way we used to work with the app's configuration retrieval to have it passed by argument instead of having the call made everywhere.

This will be especially impacting as we will be using a safe copy of the configuration later on and not have every concurrent requests work on the same one.

> Note: it should be easier to read commit by commit, I encourage you to do so 😉

### DIFF
--- a/nestor_api/lib/config.py
+++ b/nestor_api/lib/config.py
@@ -71,6 +71,7 @@ def _resolve_variable(template: str, variables: dict, path: str) -> str:
                     )
 
                 final_value = re.sub(f"{{{{{var_name}}}}}", var_value, final_value)
+        return final_value
 
     # Resolve pattern '$ENV_VAR'
     env_variables = re.findall(r"^\$(\w+)$", template)
@@ -78,6 +79,7 @@ def _resolve_variable(template: str, variables: dict, path: str) -> str:
         env_var = os.environ.get(env_variables[0])
         if env_var is not None:
             final_value = env_var
+        return final_value
 
     # Awaiting for implementation
     # -> Resolve vault definitions "!vault:xxx"
--- a/nestor_api/lib/config.py
+++ b/nestor_api/lib/config.py
@@ -26,18 +26,18 @@ def create_temporary_config_copy() -> str:
     return io.create_temporary_copy(Configuration.get_config_path(), "config")
 
 
-def get_app_config(app_name: str) -> dict:
+def get_app_config(app_name: str, config_path: str = Configuration.get_config_path()) -> dict:
     """Load the configuration of an app"""
     app_config_path = os.path.join(
-        Configuration.get_config_path(), Configuration.get_config_app_folder(), f"{app_name}.yaml",
+        config_path, Configuration.get_config_app_folder(), f"{app_name}.yaml",
     )
     if not io.exists(app_config_path):
         raise AppConfigurationNotFoundError(app_name)
 
     app_config = io.from_yaml(app_config_path)
-    environment_config = get_project_config()
+    project_config = get_project_config(config_path)
 
-    config = dict_utils.deep_merge(environment_config, app_config)
+    config = dict_utils.deep_merge(project_config, app_config)
 
     # Awaiting for implementation
     # validate configuration using nestor-config-validator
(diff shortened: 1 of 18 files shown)


nestor_api/lib/config.py
```diff
@@ -63,13 +61,23 @@ def _resolve_variable(template: str, variables: dict, path: str) -> str:
...
     # Resolve pattern '{{variable}}'
     referenced_names = re.findall(r"{{([\w\-.]+)}}", template)
-    for var_name in referenced_names:
-        var_value = variables.get(var_name)
-        if var_value is not None:
-            if not isinstance(var_value, str):
-                raise ConfigurationError(path, "Referenced variable should resolved to a string")
-
-            final_value = re.sub(f"{{{{{var_name}}}}}", var_value, final_value)
+    if len(referenced_names) > 0:
```
Olivier-Daire, line 64: Not sure that we need this extra check as we won't go into the loop below if length equals 0

nestor_api/lib/config.py
```diff
@@ -63,13 +61,23 @@ def _resolve_variable(template: str, variables: dict, path: str) -> str:
...
+                    raise ConfigurationError(
+                        path, "Referenced variable should resolved to a string"
+                    )
+
+                final_value = re.sub(f"{{{{{var_name}}}}}", var_value, final_value)
+
+    # Resolve pattern '$ENV_VAR'
+    env_variables = re.findall(r"^\$(\w+)$", template)
+    if len(env_variables) > 0:
+        env_var = os.environ.get(env_variables[0])
```
Olivier-Daire, line 78: We will never have more than one en variable?

nestor_api/lib/git.py
```diff
@@ -18,77 +17,74 @@ def branch(repository_dir, branch_name):
...
     )
 
     return repository_dir
 
 
-def get_last_tag(repository_dir):
+def get_last_tag(repository_dir: str) -> str:
     """Retrieves the last tag of a repository (locally)"""
-    return io.execute("git describe --always --abbrev=0", repository_dir)
+    return io.execute("git describe --always --abbrev=0", repository_dir).rstrip()
```
Olivier-Daire, line 36: I am wondering if we should not put `.rstrip()` directly in `io.execute()` method.
Do you think of any case where we would want to keep trailing whitespaces?

nestor_api/lib/docker.py
```diff
@@ -49,30 +41,26 @@ def build(app_name, context):
...
-    image_tag = git.get_last_tag(repository)
-
     if not has_docker_image(app_name, image_tag):
         raise RuntimeError("Docker image not available")
 
-    registry = app_config["docker"]["registry"]
+    # This will need to be done a bit differently to work with other registries (GCP)
+    # -> the config schema currently expects
+    #   {docker: {registries: {[name: string]: {id: string, organization: string}[]}}}
+    registry = app_config["docker"]["registries"]["docker.com"][0]
```
Olivier-Daire, line 63: Are we sure all those keys will always be in the dict? And that it would always have at least one value in the array?

I maybe would have the tendency to use the following to avoid raising a `KeyError` if something is missing
```python
registries = app_config.get("docker", {}).get("registries", {}).get("docker.com")
if registries is None: # we might even check for len() here, not sure it is 100% necessary, though
    logger.error("Registries not found in app_config", app_config)
    raise Exception("blabla")

registry = registries[0]
```

What do you think?

nestor_api/lib/config.py
benard-g: The only use case I encountered for this is when I need to reference a build-time variable (a `NPM_TOKEN` in my case).
The regex will match the whole content, so "xxx $VAR xxx" won't be resolved, only 1 var supported.

I don't feel like allowing some patterns like `$SUB_DOMAIN.$DOMAIN" would be very nice but it is really just my opinion

nestor_api/lib/docker.py
benard-g: I wanted to discuss if we could simplify this field as it looks a bit too complex compared to what we need. I added a note in the project's board 🙂

nestor_api/lib/git.py
benard-g: I don't see any 🤔 , I'll put it in `io.execute()` then.

PR #40, opened 2020-07-11
Remove pytest

patrickportal:
Last steps to replace **pytest** by **unittest** 
- Remove it from validation
- Remove from dependencies


Makefile
```diff
@@ -1,10 +1,10 @@
-.PHONY: help format lint mypy pytest test-all vulncheck
+.PHONY: help format lint mypy unittest test-all vulncheck
 
 help:
 	@echo "format    - Format python code with isort and black"
 	@echo "lint      - Check style with pylint"
 	@echo "mypy      - Run the static type checker"
-	@echo "pytest    - Run tests suite with pytest"
+	@echo "unittest  - Run tests suite with unittest"
```
Olivier-Daire, line 7: I am wondering if we shouldn't just call it `test`, I find it more common and it avoids being tied to a framework

tests/config_validator/test_validate.py
```diff
@@ -1,109 +1,114 @@
-# pylint: disable=missing-function-docstring,missing-module-docstring
+# pylint: disable=missing-function-docstring,missing-module-docstring,bad-continuation
```
Olivier-Daire, line 1: We now have a [`.pylintrc` for tests](https://github.com/ChauffeurPrive/nestor-api/blob/master/tests/.pylintrc#L144) that already handle those rules, except `bad-continuation` that should be avoided as much as possible. But, if necessary, I think that we are use to disabling it only for the problematic line

tests/config_validator/test_validate.py
```diff
@@ -1,109 +1,114 @@
...
-def test_validate_error_validation_target(mocker):
-    fixtures_path = Path(os.path.dirname(__file__), "..", "__fixtures__").resolve()
-    mocker.patch.object(config_validator, "build_apps_path", return_value=fixtures_path)
-    mocker.patch.object(Configuration, "get_validation_target", return_value="SOME_VALUE")
-    with pytest.raises(
-        Exception,  # pylint: disable=bad-continuation
-        match="There is no configuration to be validated. Be sure to define a valid NESTOR_VALIDATION_TARGET",  # pylint: disable=bad-continuation,line-too-long
+class TestValidateLibrary(unittest.TestCase):
+    def test_is_yaml_file(self):
+        assert not config_validator.is_yaml_file("invalid_name")
```
Olivier-Daire, line 17: It helps having better error message when running the tests 😉 
```suggestion
        self.assertEqual(not config_validator.is_yaml_file("invalid_name"), False)
```
https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertEqual

Same for the rest of the file

patrickportal: Nice, 
I removed `assert` in the test_git (last use of assert) https://github.com/ChauffeurPrive/nestor-api/pull/40/commits/cba1c1ae7eb6df40b2bad57f2ef0ab8ac36aa5ef

tests/config_validator/test_validate.py
patrickportal: I didn't added this to each line because it would have been on each line of the multiline method definition, not nice
But I removed unused pylintinstruction in the rest  of the project while  at it https://github.com/ChauffeurPrive/nestor-api/pull/40/commits/4aea78e3bba6a52c35e1e780c88a2250c15284fc

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.7
+FROM python:3.7-slim
```
patrickportal, line 1: Docker resulting image size will be **196Mio** instead of **927Mio**

patrickportal:
@Olivier-Daire `pytest-cov` has been replaced by `coverage` package (as we get rid of pytest)
I created a new command `make coverage` inorder to generate a html report, seems better to not always generate bunch of html files for each test suite.

I will fix the coverage before to merge 😉

patrickportal:
Replaced by https://github.com/ChauffeurPrive/nestor-api/pull/41

PR #42, opened 2020-07-15
Add build_api

benard-g:
Add a new endpoint to build a new image from the latest version of an application.
The new route is `POST /api/builds/:app_name`


tests/.pylintrc
```diff
@@ -143,7 +143,8 @@ disable=print-statement,
         missing-module-docstring,
         missing-function-docstring,
         no-self-use,
-        protected-access
+        protected-access,
+        too-many-arguments
```
maximejimenez, line 147: Not sure why we need this ? 🤔

nestor_api/api/api_routes/builds/build_app.py
```diff
@@ -0,0 +1,88 @@
...
+
+# pylint: disable=broad-except
+def _differed_build(app_name: str):
+    config_dir = None
+    app_dir = None
+
+    try:
+        # Retrieve app's configuration
+        config_dir = config.create_temporary_config_copy()
+        config.change_environment("staging", config_dir)
```
maximejimenez, line 22: Not sure we want to "hardcode" the staging env here, maybe **at least** have this in configuration or retrieve from workflows config in the project.yaml.
Maybe it would be clearer to configure this inside the project directly as an env variable.

tests/.pylintrc
chris-verclytte: It is because of one of the linting rule saying we don't accept more than 5 arguments per function. It is a bit strict but I think it forces to break down big functions into smaller ones enforcing clean code good practices :)

benard-g: In the test, I sometimes need a lot of argument because of the `@mocks/@patches`.
Notice that I only edited the pylintrc from the tests folder, the rule is still active in `nestor_api`

README.md
```diff
@@ -16,10 +16,11 @@ pipenv sync # Add --dev to also install dev dependencies
...
-|      `NESTOR_CONFIG_APPS_FOLDER` | `apps`                 |      | The application config folder |
-| `NESTOR_CONFIG_PROJECT_FILENAME` | `project.yaml`         |      | The project config file       |
-|           `NESTOR_PRISTINE_PATH` | `/tmp/nestor/pristine` |      | Pristine path                 |
-|               `NESTOR_WORK_PATH` | `/tmp/nestor/work`     |      | Work path                     |
+|                              Key | Default                | Unit | Comment                                                     |
+| -------------------------------: | ---------------------- | ---- | ----------------------------------------------------------- |
+|             `NESTOR_CONFIG_PATH` |                        |      | Configuration path                                          |
+|      `NESTOR_CONFIG_APPS_FOLDER` | `apps`                 |      | The application config folder                               |
+| `NESTOR_CONFIG_PROJECT_FILENAME` | `project.yaml`         |      | The project config file                                     |
+|   `NESTOR_CONFIG_DEFAULT_BRANCH` | `HEAD`                 |      | The branch to use by default when reading the configuration |
```
maximejimenez, line 24:
```suggestion
|   `NESTOR_CONFIG_DEFAULT_BRANCH` | `staging`                 |      | The branch to use by default when reading the configuration |
```

tests/api/api_routes/builds/test_build_app.py
```diff
@@ -0,0 +1,168 @@
...
+        }
+        config_mock.create_temporary_config_copy.return_value = "/tmp/config"
+        config_mock.get_app_config.return_value = app_config
+
+        git_mock.create_working_repository.return_value = "/tmp/working/repo"
+
+        app_mock.get_version.return_value = "1.0.0"
+        git_mock.tag.return_value = "1.0.0-sha-a1b2c3d4"
+
+        docker_mock.build.return_value = "my-app@1.0.0-sha-a1b2c3d4"
```
maximejimenez, line 52: Could we have this in the setup and only override needed value for error test cases ?
I would be easier to read IMO.

README.md
benard-g: oops...

PR #43, opened 2020-07-15
Add workflow lib

chris-verclytte:


nestor_api/lib/workflow.py
```diff
@@ -0,0 +1,45 @@
...
+        )
+
+    return ready_to_progress_app
+
+
+def is_app_ready_to_progress(app_dir: str, current_step: Optional[str], next_step: str) -> bool:
+    """Determines if an app can be advanced to the next step."""
+    if current_step is None:
+        return True
+    return compare_step_hashes(app_dir, current_step, next_step)
```
maximejimenez, line 36:
```suggestion
def should_app_progress(app_dir: str, current_step: Optional[str], next_step: str) -> bool:
    """Determines if an app can be advanced to the next step."""
    if current_step is None:
        return True
    return compare_step_hashes(app_dir, current_step, next_step)
```

tests/lib/test_config.py
```diff
@@ -234,3 +234,55 @@ def test_resolve_variables_deep_with_invalid_reference(self, _io_mock):
...
+        self.assertIsNone(previous_step)
+
+    def test_get_previous_step_raises_error_with_incorrect_config(self, _io_mock):
+        """Should raise an error if config is malformed."""
+        with self.assertRaises(KeyError):
+            config.get_previous_step({}, "step1")
+
+    @patch("nestor_api.lib.config.os", autospec=True)
+    def test_list_apps_config(self, os_mock, io_mock):
+        """Should return an dictionary of apps config."""
```
benard-g, line 255:
```suggestion
        """Should return a dictionary of apps config."""
```

tests/lib/test_config.py
```diff
@@ -234,3 +234,55 @@ def test_resolve_variables_deep_with_invalid_reference(self, _io_mock):
...
+            {
+                "app-1": {"name": "app-1", "config_key": "value for app-1"},
+                "app-2": {"name": "app-2", "config_key": "value for app-2"},
+                "app-3": {"name": "app-3", "config_key": "value for app-3"},
+            },
+        )
+
+    @patch("nestor_api.lib.config.os.path.isdir", autospec=True)
+    def test_list_apps_config_with_incorrect_apps_path(self, is_dir_mock, _io_mock):
+        """Should return an dictionary of apps config."""
```
benard-g, line 284:
```suggestion
        """Should return a dictionary of apps config."""
```

tests/lib/test_workflow.py
```diff
@@ -0,0 +1,95 @@
+# pylint: disable=missing-class-docstring disable=missing-function-docstring disable=missing-module-docstring
```
benard-g, line 1: You normally don't need those anymore 😉

nestor_api/lib/config.py
```diff
@@ -125,3 +126,25 @@ def __resolve(value, variables, path, errors):
         raise AggregatedConfigurationError(errors)
 
     return resolved_config
+
+
+def get_previous_step(config_object: dict, target: str) -> Optional[str]:
```
maximejimenez, line 131:
```suggestion
def get_previous_step(config: dict, target: str) -> Optional[str]:
```

nestor_api/lib/config.py
```diff
@@ -125,3 +126,25 @@ def __resolve(value, variables, path, errors):
...
+    if index > 0:
+        return config_object["workflow"][index - 1]
+    return None
+
+
+# pylint: disable=bad-continuation
+def list_apps_config(
+    config_path=Configuration.get_config_path(), apps_path=Configuration.get_config_app_folder()
+) -> dict:
+    """Retrieves all of the application names."""
```
maximejimenez, line 143:
```suggestion
    """Retrieves all of the application names and configurations."""
```

tests/lib/test_workflow.py
```diff
@@ -0,0 +1,95 @@
...
+        result = workflow.compare_step_hashes("path_to/app_dir", "step-1", "step-2")
+        self.assertEqual(git_mock.branch.call_count, 2)
+        self.assertEqual(git_mock.get_last_commit_hash.call_count, 2)
+        git_mock.branch.assert_has_calls(
+            [call("path_to/app_dir", "step-1"), call("path_to/app_dir", "step-2")]
+        )
+        git_mock.get_last_commit_hash.assert_has_calls(
+            [call("path_to/app_dir"), call("path_to/app_dir")]
+        )
+        self.assertFalse(result)
```
benard-g, line 23: Would be more readable IMO but up to you.
```suggestion
        git_mock.get_last_commit_hash.side_effect = ["hash-1", "hash-2"]
        
        result = workflow.compare_step_hashes("path_to/app_dir", "step-1", "step-2")
        
        self.assertEqual(git_mock.branch.call_count, 2)
        self.assertEqual(git_mock.get_last_commit_hash.call_count, 2)
        git_mock.branch.assert_has_calls(
            [call("path_to/app_dir", "step-1"), call("path_to/app_dir", "step-2")]
        )
        git_mock.get_last_commit_hash.assert_has_calls(
            [call("path_to/app_dir"), call("path_to/app_dir")]
        )
        self.assertFalse(result)
```

nestor_api/lib/config.py
```diff
@@ -125,3 +126,25 @@ def __resolve(value, variables, path, errors):
...
+    apps_path = os.path.join(config_path, apps_path)
+
+    if not os.path.isdir(apps_path):
+        raise ValueError(apps_path)
+    apps_files = [f for f in os.listdir(apps_path) if os.path.isfile(os.path.join(apps_path, f))]
```
maximejimenez, line 148: Are you sure that `listdir` will not throw itself an Exception here?

maximejimenez, line 148:
```suggestion
    if not os.path.isdir(apps_path):
        raise ValueError(apps_path)
    apps_files = [file for file in os.listdir(apps_path) if os.path.isfile(os.path.join(apps_path, file))]
```

tests/lib/test_config.py
```diff
@@ -234,3 +234,55 @@ def test_resolve_variables_deep_with_invalid_reference(self, _io_mock):
...
+
+    def test_get_previous_step_without_previous_step(self, _io_mock):
+        """Should answer with None as the previous step does not exist."""
+        previous_step = config.get_previous_step({"workflow": ["step1", "step2", "step3"]}, "step1")
+        self.assertIsNone(previous_step)
+
+    def test_get_previous_step_raises_error_with_incorrect_config(self, _io_mock):
+        """Should raise an error if config is malformed."""
+        with self.assertRaises(KeyError):
+            config.get_previous_step({}, "step1")
```
maximejimenez, line 251: If you move in workflow lib, don't forget to move test as well

tests/lib/test_config.py
```diff
@@ -234,3 +234,55 @@ def test_resolve_variables_deep_with_invalid_reference(self, _io_mock):
...
+            # pylint: disable=no-else-return
+            if arg == "path/to/app-1":
+                return {"name": "app-1", "config_key": "value for app-1"}
+            elif arg == "path/to/app-2":
+                return {"name": "app-2", "config_key": "value for app-2"}
+            elif arg == "path/to/app-3":
+                return {"name": "app-3", "config_key": "value for app-3"}
+            return None
+
+        io_mock.from_yaml.side_effect = yaml_side_effect
```
maximejimenez, line 270:
```suggestion
        io_mock.from_yaml.side_effect = [
          {"name": "app-1", "config_key": "value for app-1"},
          {"name": "app-2", "config_key": "value for app-2"},
          {"name": "app-3", "config_key": "value for app-3"},
        ]
```

You can also do it this way (but your solution works fine, just FYI, also you check for the params that are given, this solution does not)

tests/lib/test_workflow.py
```diff
@@ -0,0 +1,95 @@
...
+        config_mock.create_temporary_config_copy.return_value = "config-path"
+        config_mock.get_previous_step.return_value = "step-1"
+        config_mock.get_project_config.return_value = {"some": "config"}
+        config_mock.list_apps_config.return_value = {
+            "app1": {"name": "app1", "git": {"origin": "fake-remote-url-1"}},
+            "app2": {"name": "app2", "git": {"origin": "fake-remote-url-2"}},
+            "app3": {"name": "app2", "git": {"origin": "fake-remote-url-3"}},
+        }
+        git_mock.create_working_repository.return_value = "path_to/app_dir"
+        is_app_ready_to_progress_mock.return_value = True
```
maximejimenez, line 73:
```suggestion
        is_app_ready_to_progress_mock.side_effect = [True, True, False]
```

Just to avoid having only true in your final array where you expect.

nestor_api/lib/config.py
```diff
@@ -125,3 +126,25 @@ def __resolve(value, variables, path, errors):
...
+    """ Returns the previous step in the defined workflow """
+    index = config_object["workflow"].index(target)
+    if index > 0:
+        return config_object["workflow"][index - 1]
+    return None
+
+
+# pylint: disable=bad-continuation
+def list_apps_config(
+    config_path=Configuration.get_config_path(), apps_path=Configuration.get_config_app_folder()
```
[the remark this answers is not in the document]
chris-verclytte: Then it would go the same for config path, I just tried to be consistent.

nestor_api/lib/config.py
chris-verclytte: Outdated as I changed a little bit the way it is handled.

nestor_api/lib/config.py
```diff
@@ -125,3 +126,26 @@ def __resolve(value, variables, path, errors):
...
+
+    apps_config_hashmap = {}
+    for file_path in os.listdir(apps_path):
+        basename = os.path.basename(file_path)
+        filename = PurePath(basename)
+        file_extension = "".join(filename.suffixes)
+        app_name = filename.name.replace(file_extension, "")
+
+        # Prevent parsing other files than configuration ones (directories, incorrect extension)
+        if file_extension != ".yml":
```
benard-g, line 147: Also allow `.yaml`

nestor_api/lib/workflow.py
```diff
@@ -0,0 +1,49 @@
...
+    project_config = config.get_project_config()
+    previous_step = get_previous_step(project_config, next_step)
+    apps = config.list_apps_config()
+
+    ready_to_progress_app: dict = {}
+
+    for (app_name, app_config) in apps.items():
+        repository_url = app_config["git"]["origin"]
+        app_dir = git.create_working_repository(app_name, repository_url)
+        ready_to_progress_app[app_name] = should_app_progress(app_dir, previous_step, next_step)
```
benard-g, line 25: It may be possible to return `app_config` and `app_dir` along with the status because we will probably need them later on. What do you think ?
Or if we don't want to return them, maybe clean `app_dir` to free some disk space.

chris-verclytte: Remove will be enough at the moment, let's modify it only if needed in the future :)

PR #47, opened 2020-07-17
Add read/write and yaml export to io library

maximejimenez:
Add function required for kubernetes library in order to read/write configuration file and deploy using `kubectl apply`.

_Note:_ I also took the time to add missing typings to already existing function (both params and return value)

### DIFF
--- a/nestor_api/lib/io.py
+++ b/nestor_api/lib/io.py
@@ -1,19 +1,19 @@
 """I/O library"""
-from datetime import datetime
 import errno
 import math
 import os
-from pathlib import Path
-from random import random
 import shutil
 import subprocess
+from datetime import datetime
+from pathlib import Path
+from random import random
 
 import yaml
 
 from nestor_api.config.config import Configuration
 
 
-def copy(source, destination):
+def copy(source: str, destination: str) -> None:
     """Copy file or directory with its content from source to destination"""
     try:
         shutil.copytree(source, destination)
@@ -24,7 +24,7 @@ def copy(source, destination):
             raise
 
 
-def create_temporary_directory(directory_prefix=""):
+def create_temporary_directory(directory_prefix: str = "") -> str:
     """Creates a temporary directory with an optional prefix"""
     tmp_directory_path = get_temporary_directory_path(directory_prefix)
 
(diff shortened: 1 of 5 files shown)


nestor_api/lib/io.py
```diff
@@ -86,12 +86,23 @@ def get_temporary_directory_path(prefix=""):
...
-def remove(file_path):
+def to_yaml(data: dict) -> str:
+    """Converts a dictionary into a valid yaml string"""
+    return yaml.safe_dump(data)
+
+
+def read(file_path: str) -> str:
+    """Read the file content at the given path"""
+    file = open(file_path, "r")
+    return file.read()
```
benard-g, line 102: You forgot to call `file.close()`.
I recommend to use the `with` structure to ensure this
```python
def read(file_path: str) -> str:
    """Read the file content at the given path"""
    with open(file_path, "r") as file:
        file_content = file.read()
    return file_content
```

nestor_api/lib/io.py
```diff
@@ -34,46 +34,46 @@ def create_temporary_directory(directory_prefix=""):
...
 
 
-def exists(file_path):
+def exists(file_path: str) -> bool:
     """Checks if a file exists"""
     return Path(file_path).exists()
 
 
-def from_yaml(file_path):
+def from_yaml(file_path: str) -> dict:
```
benard-g, line 55: What do you think of improving this function's name while you're at it ? (I find the `io.from_yaml()` a bit weird personally).
Maybe something similar to "parse_yaml" or "read_yaml" (and then the `to_yaml` you added becomes "write_yaml").

nestor_api/lib/io.py
```diff
@@ -100,3 +111,9 @@ def remove(file_path):
             os.remove(file_path)
         else:
             raise
+
+
+def write(file_path: str, content: str) -> None:
+    """ Write the string content into the file at the given path"""
+    file = open(file_path, "w")
+    file.write(content)
```
benard-g, line 119: Same comment as above about closing and the `with` structure.

Also, you may need to use `"w+"` to create the file if it does not exists (not a 100% sure on this one).

maximejimenez: From what I read, I can see that `+` is only to read as well when opening with w (for writing).

nestor_api/lib/io.py
```diff
@@ -86,12 +83,27 @@ def get_temporary_directory_path(prefix=""):
...
+    """Read the file content at the given path"""
+    with open(file_path, "r") as file:
+        file_content = file.read()
+    return file_content
+
+
+def read_yaml(file_path: str) -> dict:
+    """Returns a dictionary from yaml file path"""
+    with open(file_path, "r") as file_data:
+        yaml_data = file_data.read()
```
benard-g, line 101: You could call the above `read` function here, do as you prefer.

PR #48, opened 2020-07-20
Applications and cronjobs getters for an application configuration

maximejimenez:
Add the ability to retrieve `applications` and `cronjobs` for an application configuration in order to create one deployment for each process.

### DIFF
--- a/nestor_api/lib/config.py
+++ b/nestor_api/lib/config.py
@@ -46,6 +46,16 @@ def get_app_config(app_name: str, config_path: str = Configuration.get_config_pa
     return _resolve_variables_deep(config)
 
 
+def get_cronjobs(app_config: dict) -> list:
+    """Get cronjobs processes"""
+    return [process for process in app_config["processes"] if process["is_cronjob"]]
+
+
+def get_processes(app_config: dict) -> list:
+    """Get processes (not cronjobs)"""
+    return [process for process in app_config["processes"] if not process["is_cronjob"]]
+
+
 def get_project_config(config_path: str = Configuration.get_config_path()) -> dict:
     """Load the global configuration of the project"""
     project_config_path = os.path.join(config_path, Configuration.get_config_project_filename())
--- a/tests/lib/test_config.py
+++ b/tests/lib/test_config.py
@@ -88,6 +88,42 @@ class TestConfigLibrary(unittest.TestCase):
 
         self.assertEqual("Configuration file not found for app: some-app", str(context.exception))
 
+    def test_get_processes(self, _io_mock):
+        app_config = {
+            "name": "my-app",
+            "processes": [
+                {"name": "web", "start_command": "./web", "is_cronjob": False,},
+                {"name": "cleaner", "start_command": "./clean", "is_cronjob": True,},
+                {"name": "worker", "start_command": "./worker", "is_cronjob": False,},
+            ],
+        }
+
+        applications = config.get_processes(app_config)
+
+        self.assertEqual(
+            applications,
+            [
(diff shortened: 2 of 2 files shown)


nestor_api/lib/config.py
```diff
@@ -46,6 +46,16 @@ def get_app_config(app_name: str, config_path: str = Configuration.get_config_pa
     return _resolve_variables_deep(config)
 
 
+def get_applications(app_config: dict) -> list:
```
chris-verclytte, line 49: According to the docstring, I would rather name the function `get_application_processes`

nestor_api/lib/config.py
```diff
@@ -46,6 +46,16 @@ def get_app_config(app_name: str, config_path: str = Configuration.get_config_pa
...
+    """Get applications processes (not cronjobs)"""
```
chris-verclytte, line 50:
```suggestion
    """Get application processes (not cronjobs)"""
```

nestor_api/lib/config.py
```diff
@@ -46,6 +46,16 @@ def get_app_config(app_name: str, config_path: str = Configuration.get_config_pa
...
+    return list(filter(lambda process: process["is_cronjob"] is False, app_config["processes"]))
```
chris-verclytte, line 51: What about comprehension here like: 
```python
return [process for process in app_config["processes"] if process["is_cronjob"] is False]
```
I find it clearer to read than function nesting with lambda

nestor_api/lib/config.py
```diff
@@ -46,6 +46,16 @@ def get_app_config(app_name: str, config_path: str = Configuration.get_config_pa
...
+
+
+def get_cronjobs(app_config: dict) -> list:
```
chris-verclytte, line 54: Then either `get_application_cronjobs` or rename the first one ``get_processes``

nestor_api/lib/config.py
benard-g, line 51: You could use a list comprehension here 🙂 
```suggestion
    return [process for process in app_config["processes"] if not process["is_cronjob"]]
```

nestor_api/lib/config.py
```diff
@@ -46,6 +46,16 @@ def get_app_config(app_name: str, config_path: str = Configuration.get_config_pa
...
+    """Get cronjobs processes"""
+    return list(filter(lambda process: process["is_cronjob"] is True, app_config["processes"]))
```
benard-g, line 56: Same here
```suggestion
    return [process for process in app_config["processes"] if process["is_cronjob"]]
```

nestor_api/lib/config.py
```diff
@@ -46,6 +46,16 @@ def get_app_config(app_name: str, config_path: str = Configuration.get_config_pa
     return _resolve_variables_deep(config)
 
 
+def get_cronjobs(app_config: dict) -> list:
+    """Get cronjobs processes"""
+    return [process for process in app_config["processes"] if process["is_cronjob"]]
```
chris-verclytte, line 51: Maybe 
```suggestion
    return [cronjob for cronjob in app_config["processes"] if process["is_cronjob"]]
```

tests/lib/test_config.py
```diff
@@ -88,6 +88,42 @@ def test_get_app_config_when_not_found(self, configuration_mock, io_mock):
 
         self.assertEqual("Configuration file not found for app: some-app", str(context.exception))
 
+    def test_get_processes(self, _io_mock):
+        app_config = {
```
chris-verclytte, line 92:
```suggestion
        APP_CONFIG = {
```

PR #49, opened 2020-07-20
Disallow duplicates in yaml files

Kenny407:
From the investigation I did we have two options, either replace completely the pyyaml library which is already used in the project or address the workaround to detect and handle duplicated keys in yaml files.

The main reason of detecting duplicated it's to provide as much feedback as possible to developers that use Nestor. For instance, if they misconfigure Nestor, we would be able to provide the correct feedback of which variable is duplicated.

Also I followed the workarounds of other devs to understand whether it was more feasible to remove it or keep a workaround and then remove it if it's addressed by the maintainers or go with https://pypi.org/project/ruamel.yaml/ for instance, but overall third party libs are a bit limited in terms of loading yamls and managing them, so we would lose some features as merging yaml files programmatically, and so on.


yaml_lib/duplicate_keys.py
```diff
@@ -0,0 +1,47 @@
+"""Constructor to not allow duplicated keys
```
benard-g, line 1:
```suggestion
"""A custom YAML loader complying with the YAML specification.
```

tests/lib/test_duplicate_keys.py
```diff
@@ -0,0 +1,20 @@
...
+class TestDetectDuplicateKeysInYaml(TestCase):
+    def test_read_yaml_disallow_duplicate_keys(self):
+        """Assert that read_yaml disallows duplicate keys"""
+        yaml_fixture_path = Path(
+            os.path.dirname(__file__), "..", "__fixtures__", "yaml", "duplicated_keys.yaml"
+        ).resolve()
+
+        with open(yaml_fixture_path, "r") as file_data:
+            yaml_data = file_data.read()
+            with self.assertRaises(Exception):
```
benard-g, line 19: Since the objective is to have accurate error reporting, could you perhaps have a more specific assertion (type + message) ? 🙂

tests/lib/test_duplicate_keys.py
```diff
@@ -0,0 +1,20 @@
...
+from pathlib import Path
+from unittest import TestCase
+
+import yaml
+
+from yaml_lib.duplicate_keys import YamlLoader  # type: ignore
+
+
+class TestDetectDuplicateKeysInYaml(TestCase):
+    def test_read_yaml_disallow_duplicate_keys(self):
```
benard-g, line 11: Maybe also add another test with a `success` outcome (no duplicates) ?

Kenny407:
@benard-g thanks for the comments and the suggestions! I took your ideas and added them. I hope the library it's more comfortable to use. I made the changes in the last commit: https://github.com/ChauffeurPrive/nestor-api/pull/49/commits/00d6759a034df65a8ae86aabe5c247ea97603211 

I came up with more descriptive names to avoid any misunderstanding. Let me know what you think ! and thanks in advance

yaml_lib/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""Library to handle all the YAML related functions"""
+
+from yaml_lib.load_yaml import load_yaml_from_path
```
benard-g, line 3: I think you could also write, but it really don't change anything
```suggestion
from .load_yaml import load_yaml_from_path
```

yaml_lib/duplicate_keys_loader.py
```diff
@@ -14,23 +14,23 @@
 import yaml
 
 
-class YamlLoader(yaml.SafeLoader):  # pylint: disable=too-many-ancestors
-    """Custom YAML Loader to comply with the official YAML Spec"""
+class DuplicateKeysLoader(yaml.SafeLoader):  # pylint: disable=too-many-ancestors
```
benard-g, line 17: I actually preferred `YamlLoader` 😅

yaml_lib/load_yaml.py
```diff
@@ -0,0 +1,19 @@
...
+
+def load_yaml_from_path(file_path: str) -> dict:
+    """Loads a yaml file from a path using the DuplicateKeysLoader
+
+  Args:
+      file_path (str): The file path
+
+  Returns:
+      dict: The representation of the yaml file as a dict
+  """
```
benard-g, line 16:
```suggestion
    """Loads a yaml file from a path using the DuplicateKeysLoader

    Args:
          file_path (str): The file path

    Returns:
          dict: The representation of the yaml file as a dict
    """
```

nestor_api/lib/io.py
```diff
@@ -98,10 +98,7 @@ def read(file_path: str) -> str:
 
 def read_yaml(file_path: str) -> dict:
     """Returns a dictionary from yaml file path"""
-    with open(file_path, "r") as file_data:
-        yaml_data = file_data.read()
-
-    return yaml.load(yaml_data, Loader=YamlLoader)
+    return yaml_lib.load_yaml_from_path(file_path)
```
benard-g, line 101: 👍

PR #53, opened 2020-07-21
Add kubectl library

benard-g:
Add the `kubectl` function that will allow us to retrieve the resources configurations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,3 +27,4 @@ pipenv sync # Add --dev to also install dev dependencies
 |    `NESTOR_PROBES_DEFAULT_DELAY` | `30`                   | `seconds` | Default delay for probes if not configured                  |
 |   `NESTOR_PROBES_DEFAULT_PERIOD` | `10`                   | `seconds` | Default period for probes if not configured                 |
 |  `NESTOR_PROBES_DEFAULT_TIMEOUT` | `1`                    | `seconds` | Default timeout for probes if not configured                |
+|          `NESTOR_K8S_HTTP_PROXY` |                        | `URL`     | The kubernetes HTTP_PROXY                                   |
--- a/README.md
+++ b/README.md
@@ -27,4 +27,4 @@ pipenv sync # Add --dev to also install dev dependencies
 |    `NESTOR_PROBES_DEFAULT_DELAY` | `30`                   | `seconds` | Default delay for probes if not configured                  |
 |   `NESTOR_PROBES_DEFAULT_PERIOD` | `10`                   | `seconds` | Default period for probes if not configured                 |
 |  `NESTOR_PROBES_DEFAULT_TIMEOUT` | `1`                    | `seconds` | Default timeout for probes if not configured                |
-|          `NESTOR_K8S_HTTP_PROXY` |                        | `URL`     | The kubernetes HTTP_PROXY                                   |
+|          `NESTOR_K8S_HTTP_PROXY` |                        |           | The kubernetes HTTP_PROXY                                   |
--- a/nestor_api/config/kubernetes.py
+++ b/nestor_api/config/kubernetes.py
@@ -1,0 +1,12 @@
+"""Nestor-api kubernetes configuration"""
+import os
+
+
+# pylint: disable=too-few-public-methods
+class KubernetesConfiguration:
+    """Kubernetes configuration"""
+
+    @staticmethod
+    def get_http_proxy() -> str:
+        """Returns the kubernetes http proxy"""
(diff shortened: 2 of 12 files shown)


nestor_api/config/kubernetes.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+# pylint: disable=too-few-public-methods
+class KubernetesConfiguration:
+    """Kubernetes configuration"""
+
+    @staticmethod
+    def get_http_proxy() -> str:
+        """Returns the kubernetes http proxy"""
+        return os.environ["NESTOR_K8S_HTTP_PROXY"]
```
maximejimenez, line 12: Also the file name.

I think we should be consistent between k8s and kubernetes, here the variables has `K8S`, builders are in a `k8s` folder, we could change to `kubernetes` but just to have same naming would be easier I think

```suggestion
"""Nestor-api k8s configuration"""
import os


# pylint: disable=too-few-public-methods
class K8sConfiguration:
    """K8s configuration"""

    @staticmethod
    def get_http_proxy() -> str:
        """Returns the k8s http proxy"""
        return os.environ["NESTOR_K8S_HTTP_PROXY"]
```

tests/config/test_kubectl.py
```diff
@@ -0,0 +1,17 @@
...
+class TestKubernetesConfig(TestCase):
+    @patch.dict(os.environ, {"NESTOR_K8S_HTTP_PROXY": "k8s-proxy.my-domain.com"})
+    def test_get_http_proxy(self):
+        self.assertEqual(KubernetesConfiguration.get_http_proxy(), "k8s-proxy.my-domain.com")
+
+    @patch.dict(os.environ, {"NESTOR_K8S_HTTP_PROXY": ""})
+    def test_get_http_proxy_not_existing(self):
+        del os.environ["NESTOR_K8S_HTTP_PROXY"]
+        with self.assertRaises(Exception):
+            KubernetesConfiguration.get_http_proxy()
```
maximejimenez, line 17:
```suggestion
        with self.assertRaises(KeyError):
            KubernetesConfiguration.get_http_proxy()
```

nestor_api/config/kubernetes.py
maximejimenez, line 12: Add this in the README 🙏

nestor_api/lib/kubectl.py
```diff
@@ -0,0 +1,44 @@
...
+import nestor_api.lib.io as io
+
+
+class KubernetesResourceType(Enum):
+    """An enum representing the different resources kinds"""
+
+    DEPLOYMENT = "deployment"
+
+    def __str__(self):
+        return str(self.value)
```
maximejimenez, line 17: Maybe we could define this under `k8s` in a dedicated file.
Also, the file `kubectl` could be `cli.py` under `k8s`, what do you think, it is just to avoid spreading all k8s logic in the project.

tests/lib/test_kubectl.py
```diff
@@ -0,0 +1,30 @@
...
+        io_mock.execute.assert_called_once_with(
+            (
+                "kubectl "
+                "--context cluster "
+                "--namespace namespace "
+                "get deployment "
+                "--output=json "
+                "--selector app=app"
+            ),
+            env={**os.environ, "HTTP_PROXY": "k8s-proxy.my-domain.com",},
```
maximejimenez, line 29:
```suggestion
            env={**os.environ, "HTTP_PROXY": "k8s-proxy.my-domain.com"},
```

nestor_api/config/kubernetes.py
benard-g: I will go with `k8s` then

README.md
```diff
@@ -27,3 +27,4 @@ pipenv sync # Add --dev to also install dev dependencies
 |    `NESTOR_PROBES_DEFAULT_DELAY` | `30`                   | `seconds` | Default delay for probes if not configured                  |
 |   `NESTOR_PROBES_DEFAULT_PERIOD` | `10`                   | `seconds` | Default period for probes if not configured                 |
 |  `NESTOR_PROBES_DEFAULT_TIMEOUT` | `1`                    | `seconds` | Default timeout for probes if not configured                |
+|          `NESTOR_K8S_HTTP_PROXY` |                        | `URL`     | The kubernetes HTTP_PROXY                                   |
```
maximejimenez, line 30:
```suggestion
|          `NESTOR_K8S_HTTP_PROXY` |                        |      | The kubernetes HTTP_PROXY                                   |
```

PR #54, opened 2020-07-21
Add get_image_name & get_secret_variables in k8s builders

maximejimenez:
Add more function in k8s builders.

### DIFF
--- a/nestor_api/lib/k8s/builders.py
+++ b/nestor_api/lib/k8s/builders.py
@@ -47,7 +47,7 @@ def get_anti_affinity_zone(app_config: dict, process_name: str, templates: dict)
     return anti_affinity_zone
 
 
-def get_image_name(app_config, process_name, options):
+def get_image_name(app_config, options):
     """Returns the definitive image name"""
     image_tag = options["tag"]
     branch = options.get("branch")
@@ -55,9 +55,7 @@ def get_image_name(app_config, process_name, options):
     if branch is not None:
         image_tag = branch
 
-    app, _, _ = get_sanitized_names(app_config, process_name)
-
-    return docker.get_registry_image_tag(app, image_tag, app_config["registry"])
+    return docker.get_registry_image_tag(app_config["app"], image_tag, app_config["registry"])
 
 
 def get_probes(probes_config: dict, port: int):
--- a/nestor_api/lib/k8s/builders.py
+++ b/nestor_api/lib/k8s/builders.py
@@ -47,14 +47,9 @@ def get_anti_affinity_zone(app_config: dict, process_name: str, templates: dict)
     return anti_affinity_zone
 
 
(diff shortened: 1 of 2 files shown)


nestor_api/lib/k8s/builders.py
```diff
@@ -92,7 +106,24 @@ def get_sanitized_names(app_config: dict, process_name: str):
     return app, sanitized_process_name, metadata_name
 
 
-def get_variables(app_config: dict):
+def get_secret_variables(app_config) -> dict:
```
benard-g, line 107: Should this be inside of `lib.config` ?
(also, don't forget the typing)

nestor_api/lib/k8s/builders.py
```diff
@@ -46,6 +47,19 @@ def get_anti_affinity_zone(app_config: dict, process_name: str, templates: dict)
     return anti_affinity_zone
 
 
+def get_image_name(app_config, process_name, options):
+    """Returns the definitive image name"""
+    image_tag = options["tag"]
+    branch = options.get("branch")
+
+    if branch is not None:
+        image_tag = branch
```
benard-g, line 56: What about ?
```suggestion
    image_tag = options["branch"] if "branch" in options else options["tag"]
```
Or you could also change the function's parameter to receive `ref` directly

nestor_api/lib/k8s/builders.py
```diff
@@ -46,6 +47,19 @@ def get_anti_affinity_zone(app_config: dict, process_name: str, templates: dict)
     return anti_affinity_zone
 
 
+def get_image_name(app_config, process_name, options):
```
benard-g: typings

nestor_api/lib/k8s/builders.py
maximejimenez: I would avoid this because we are re-formatting to k8s format here to have a deployment to apply after, so it is not very generic but more k8s oriented getters.

tests/lib/k8s/test_builders.py
```diff
@@ -129,6 +129,20 @@ def test_get_anti_affinity_zone_enabled_for_process(self, parse_yaml_mock):
         )
         self.assertEqual(anti_affinity_zone, k8s_fixtures.anti_affinity_zone)
 
+    def test_get_image_name_branch(self):
+        """Returns the URL of the docker image if tag and branch are provided"""
+        app_config = {"app": "app", "registry": {"organization": "my-organization"}}
+        image_url = k8s_builders.get_image_name(
+            app_config, {"branch": "feature/api", "tag": "1.0.0-sha-a2b3c4"}
```
benard-g, line 136: You don't really need `tag` anymore, but it is still good to assert that `branch` is taken first if present

PR #55, opened 2020-07-22
Report errors when validation fails

Kenny407:
This PR aims to provide useful feedback to nestor users when they are configuring apps or projects. A list of errors will be provided if there were issues like duplicate keys.

To the contrary of what AJV does in NodeJS, jsonschema only provides the first error that it's found while validating a file against a schema, so at most we will be only able to provide one error .

### DIFF
--- a/tests/config_validator/test_validate.py
+++ b/tests/config_validator/test_validate.py
@@ -6,6 +6,7 @@ import unittest
 from unittest.mock import patch
 
 from jsonschema import ValidationError  # type: ignore
+from yaml.constructor import ConstructorError
 
 from tests.__fixtures__.example_schema import EXAMPLE_SCHEMA  # type: ignore
 from validator.errors.errors import InvalidTargetPathError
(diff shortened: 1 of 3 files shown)


validator/validate.py
```diff
@@ -62,13 +63,17 @@ def validate_deployment_files():
     Raises:
         Exception: If the configuration path has not been configured.
         Exception: If the configuration path does not exist
+
+    Returns:
+        list: A list with errors, if empty there were no errors validating the files
```
benard-g, line 68: Don't forget to add the corresponding type hint in the function's signature

tests/config_validator/test_validate.py
```diff
@@ -109,6 +111,21 @@ def test_validate_projects(
...
-            config_validator.validate_deployment_files()
+            result = config_validator.validate_deployment_files()
+            self.assertEqual(result, [])
         except Exception as err:  # pylint: disable=broad-except
             self.fail(f"validate_deployment_files() raised an unexpected Exception {err}")
+
+    @patch("validator.validate.build_apps_path")
+    @patch("validator.config.config.Configuration.get_validation_target")
+    def test_validate_apps_with_errors(self, get_validation_target_mock, build_apps_path_mock):
+        real_config_fixture_path = Path(
```
benard-g, line 122: Eventually
```suggestion
        config_fixture_path = Path(
```

validator/validate.py
Kenny407: ahhh forgot about the typings! thanks 🤗

PR #56, opened 2020-07-22
Create simple logger with basicConfig

Kenny407:
This PR adds the implementation of a very simple logger module. 

It internally wraps the library `logging` from Python. It's up to discussion whether we add more features to the logger module such as:

- Collection of logs into files and a periodical rotation (daily, weekly, etc)
- Add more levels such as: `critical` used in some projects and already existing in the logging library
- Overriding the basic config?

### Usage with context

```py
Logger.info(message="hello", context={"person": "guillaume"})
```

```sh
INFO-22-Jul-20 16:06:46: hello {"person": "guillaume"}
```

### Usage without context

```py
Logger.info("hello")
```

```sh
INFO-22-Jul-20 16:06:46: hello
```

### DIFF
--- a/nestor_api/utils/logger.py
+++ b/nestor_api/utils/logger.py
@@ -1,21 +1,66 @@
 """A simple logger implementation awaiting for proper implementation"""
 
+import json
+import logging
+
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(levelname)s-%(asctime)s: %(message)s",
+    datefmt="%d-%b-%y %H:%M:%S",
+)
+
 
 class Logger:
-    """A fake logger module to be able to collect logs later on"""
+    """A logger module"""
 
     @staticmethod
-    def debug(context: dict, message: str):
-        """Produce a log with debug level"""
+    def debug(context=None, message=""):
+        """Produce a log with debug level
+
+        Args:
+            context (dict): An object with relevant information for debugging
+            message (str): Additional message for context/debugging
+        """
+        if context is None:
+            logging.debug(message)
+        else:
+            logging.debug("%s %s", message, json.dumps(context))
 
     @staticmethod
-    def error(context: dict, message: str):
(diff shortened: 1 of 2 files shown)


nestor_api/utils/logger.py
```diff
@@ -1,21 +1,66 @@
...
+
 
 class Logger:
-    """A fake logger module to be able to collect logs later on"""
+    """A logger module"""
 
     @staticmethod
-    def debug(context: dict, message: str):
-        """Produce a log with debug level"""
+    def debug(context=None, message=""):
```
benard-g, line 17: Do you think it would be better to have the message first ? (and non-optional) so that we could write `Logger.info("message without context")`.
Or disallow the default values and always provide both when calling the logger ?

> I realize switching the arguments might not be compatible with the already written messages that use the current order (context, message)

benard-g: Also --> don't forget to add the typings 😉

tests/utils/test_logger.py
```diff
@@ -0,0 +1,57 @@
...
+    def test_logger_error_with_no_context(self, logging_mock):
+        Logger.error(message="Found user")
+        logging_mock.error.assert_called_once_with("Found user", exc_info=True)
+
+    @patch("nestor_api.utils.logger.logging", autospec=True)
+    def test_logger_error_with_exception(self, logging_mock):
+        Logger.error({"user_id": 1234}, "Found user")
+        logging_mock.error.assert_called_once_with(
+            "%s %s", "Found user", '{"user_id": 1234}', exc_info=True
+        )
```
benard-g, line 57: There is no exception provided here right ? (or did I miss something ?)

nestor_api/utils/logger.py
Kenny407: hihi, added the typings. I always forget about it. We agreed only for the prototype of the function right?

PR #57, opened 2020-07-22
Add setters (namespace, replicas & resources) in k8s builders

maximejimenez:
Add setters for resources, namespace and replicas in k8s builders.

### DIFF
--- a/nestor_api/lib/k8s/builders.py
+++ b/nestor_api/lib/k8s/builders.py
@@ -3,6 +3,7 @@
 from typing import Optional
 
 from nestor_api.config.probes import ProbesDefaultConfiguration
+from nestor_api.config.replicas import ReplicasDefaultConfiguration
 import nestor_api.lib.docker as docker
 import yaml_lib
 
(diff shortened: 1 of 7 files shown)


nestor_api/lib/k8s/builders.py
```diff
@@ -130,3 +131,67 @@ def get_variables(app_config: dict) -> dict:
...
+
+
+# pylint: disable=bad-continuation
+def set_replicas(
+    app_config: dict, process_name: str, recipe: dict, templates: dict
+) -> Optional[dict]:
+    """Attach the correct Replicas values to the recipe object and return the corresponding
+    `hpa` if applicable"""
+    scales = app_config["scales"]
+    scales_config = scales[process_name] if process_name in scales else scales["default"]
```
benard-g, line 156: Maybe ?
```suggestion
    scales_config = scales[process_name] if process_name in scales else scales.get("default", {})
```
so that it works with the default values if the `default` field is not defined

nestor_api/lib/k8s/builders.py
```diff
@@ -130,3 +131,67 @@ def get_variables(app_config: dict) -> dict:
...
+        "minReplicas": min_replicas,
+        "maxReplicas": max_replicas,
+        "targetCPUUtilizationPercentage": target_cpu_utilization_percentage,
+        **app_config.get("templateVars", {}),
+    }
+
+    return yaml_lib.parse_yaml(templates["hpa"](template_vars))
+
+
+def set_resources(app_config, process_name, recipe) -> None:
```
benard-g, line 189:
```suggestion
def set_resources(app_config: dic, process_name: str, recipe: dict) -> None:
```

tests/lib/k8s/test_builders.py
```diff
@@ -9,11 +9,38 @@ class TestK8sBuilders(TestCase):
     def compiled_templates_mock(self):
         """To mock compiled templates (with handlebars)"""
 
-        def compiled_templates(parameters):
+        def check_anti_affinity_node_template(parameters):
             self.assertEqual(parameters, {"app": "app", "process": "web"})
-            return {"mock": "template"}
+            return {"mock: template_anti_affinity_node"}
```
benard-g, line 14: Do you want a set here ? 🙂  (same for all templates)
```suggestion
            return {"mock": template_anti_affinity_node}
```
```suggestion
            return "mock: template_anti_affinity_node"
```

tests/lib/k8s/test_builders.py
```diff
@@ -277,22 +304,281 @@ def test_get_secret_variables(self):
...
+
+        app_config = {
+            "app": "app",
+            "scales": {
+                "default": {
+                    "minReplicas": 3,
+                    "maxReplicas": 9,
+                    "targetCPUUtilizationPercentage": 80,
+                },
+                "web": {"minReplicas": 5, "maxReplicas": 15, "targetCPUUtilizationPercentage": 90,},
```
benard-g, line 410: It would be nice if `black` automatically cleaned those
```suggestion
                "web": {"minReplicas": 5, "maxReplicas": 15, "targetCPUUtilizationPercentage": 90},
```

nestor_api/lib/k8s/builders.py
maximejimenez: We want this to throw because it is mandatory, I can do a custom error if you think it is better

PR #58, opened 2020-07-22
Add Git provider abstraction layer and Github implementation

chris-verclytte:
Add abstraction layer for Git provider through AbstractGitProvider.
Add implementation of this abstraction for GitHub.


tests/adapters/git/test_github_git_provider.py
```diff
@@ -0,0 +1,150 @@
...
+    def test_get_branch(self, github_mock):
+        """Should get the branch data."""
+        fake_branch_result = MagicMock(spec=Branch.Branch)
+        instance = github_mock.return_value
+        instance.get_repo.return_value.get_branch.return_value = fake_branch_result
+        github_provider = GitHubGitProvider()
+        result = github_provider.get_branch("organization", "fake-project", "fake-branch")
+        instance.get_repo.assert_called_once_with("organization/fake-project")
+        instance.get_repo.return_value.get_branch.assert_called_once_with(branch="fake-branch")
+        self.assertEqual(result, fake_branch_result)
```
benard-g, line 49: Could you format the tests this way ? 🙏 🙂 
```suggestion
        fake_branch_result = MagicMock(spec=Branch.Branch)
        instance = github_mock.return_value
        instance.get_repo.return_value.get_branch.return_value = fake_branch_result

        github_provider = GitHubGitProvider()
        result = github_provider.get_branch("organization", "fake-project", "fake-branch")

        instance.get_repo.assert_called_once_with("organization/fake-project")
        instance.get_repo.return_value.get_branch.assert_called_once_with(branch="fake-branch")
        self.assertEqual(result, fake_branch_result)
```

nestor_api/adapters/git/abstract_git_provider.py
```diff
@@ -0,0 +1,31 @@
+"""Abstract adapter for git provider."""
+
+from abc import ABCMeta, abstractmethod
+
+
+class AbstractGitProvider(metaclass=ABCMeta):
```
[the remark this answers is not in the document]
chris-verclytte: Hmmm I don't really agree on this one, I think the implementation clearly makes it more obvious and allow to easily test in unit tests.

tests/adapters/git/test_abstract_git_provier.py
```diff
@@ -0,0 +1,33 @@
...
+from unittest.mock import patch
+
+from nestor_api.adapters.git.abstract_git_provider import AbstractGitProvider
+
+
+@patch.object(AbstractGitProvider, "__abstractmethods__", set())
+class TestAbstractGitProvider(TestCase):
+    def test_get_user_info(self):
+        """Should raise a NotImplementedError."""
+        git_provider = AbstractGitProvider()
```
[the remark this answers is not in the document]
chris-verclytte: Get a look at the patch on top of the class ;)

nestor_api/adapters/git/abstract_git_provider.py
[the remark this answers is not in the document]
chris-verclytte: Oh I add not seen the difference between subclassing and using the meta interface. Don't think it changes that much but did the change :)

nestor_api/adapters/git/github_git_provider.py
```diff
@@ -0,0 +1,68 @@
...
+            if branch is None:
+                raise GitProviderError("Branch not found")
+            branch.edit_protection(enforce_admins=False, user_push_restrictions=[user_login])
+        except GithubException as err:
+            raise _format_error(err)
+
+
+def _format_error(error: GithubException):
+    message = f"GitHub operation failed with status '{error.status}' because: {error.data}"
+    return GitProviderError(message)
```
benard-g, line 68: 👍

tests/config/test_config.py
```diff
@@ -42,5 +42,5 @@ def test_get_config_project_filename_default(self):
     def test_get_config_default_branch_configured(self):
         self.assertEqual(Configuration.get_config_default_branch(), "master")
 
-    def test_get_config_default_branch_default(self):
+    def test_get_config_default_branch_daefault(self):
```
benard-g, line 45:
```suggestion
    def test_get_config_default_branch_default(self):
```

tests/config/test_git.py
```diff
@@ -0,0 +1,15 @@
+import os
+from unittest import TestCase
+from unittest.mock import patch
+
+from nestor_api.config.git import GitConfiguration
+
+
+# pylint: disable=too-few-public-methods
```
Kenny407, line 8: Maybe we could add this rule to the `.pylintrc` of the tests?

chris-verclytte: I don't think it should be, actually the rule is right, having a class with only one method would makes no sense (better to do a simple function) but the thing is that here, we know it will probably evolve with other methods :)

PR #63, opened 2020-07-27
Add workflow initialization in workflow lib

chris-verclytte:


nestor_api/api/api_routes/workflow/workflow.py
```diff
@@ -0,0 +1,60 @@
...
+    """Initialize workflow of an application."""
+    Logger.debug(
+        {"app": app, "organization": organization},
+        "[/api/workflow/init/:org/:app] Workflow initialization started",
+    )
+
+    report = None
+    try:
+        # Retrieve project configuration
+        project_config = config_lib.get_project_config()
```
benard-g, line 21: Be careful here, you should first create a copy of the configuration to ensure that no concurrent request updates the config while you use it.
Have a look at what is made [in the build api](https://github.com/ChauffeurPrive/nestor-api/blob/master/nestor_api/api/api_routes/builds/build_app.py#L22-L24)

It should look like this :
```python
config_dir = config_lib.create_temporary_config_copy()
config_lib.change_environment(Configuration.get_config_default_branch(), config_dir)
app_config = config_lib.get_app_config()
```

This makes me raise, once again, a question you had before :
- what should the default branch be ? how should we determine it ?

nestor_api/api/api_routes/workflow/workflow.py
```diff
@@ -0,0 +1,60 @@
...
+            {"organization": organization, "app": app, "report": report, "message": message},
+            status,
+        )
+    # pylint: disable=broad-except
+    except Exception as err:
+        Logger.error(
+            {"organization": organization, "app": app, "report": report, "err": str(err)},
+            "[/api/workflow/init/:org/:app] Workflow initialization failed",
+        )
+        print(traceback.print_exc())
```
benard-g, line 50: The Logger should now handle the printing of the stacktrace since https://github.com/ChauffeurPrive/nestor-api/pull/56.
Of course, don't hesitate to double-check in case I mixed things up here 🙏 
```suggestion
        Logger.error(
            {"organization": organization, "app": app, "report": report},
            "[/api/workflow/init/:org/:app] Workflow initialization failed",
        )
```

nestor_api/lib/workflow.py
```diff
@@ -49,3 +52,92 @@ def get_previous_step(project_config: dict, target: str) -> Optional[str]:
...
+    app_config = config.get_app_config(app_name)
+    workflow_from_conf = app_config["workflow"] if app_config else []
+    master_tag = GitConfiguration.get_master_tag()
+
+    workflow_branches = [
+        branch_name for branch_name in workflow_from_conf if branch_name != master_tag
+    ]
+
+    branches = {}
+    status = "success"
```
benard-g, line 78: An enum for the status could be a plus, but not required 🙂

nestor_api/lib/workflow.py
```diff
@@ -49,3 +52,92 @@ def get_previous_step(project_config: dict, target: str) -> Optional[str]:
...
+    """Initialize workflow for a given repo by creating all workflow branches.
+    This function is idempotent which means it will not try to recreate a
+    branch that already exist. However if a branch already exist but is not protected,
+    it will be set to protected."""
+
+    # Switch to staging environment in order to get application configuration
+    config.change_environment("staging")
+
+    # Get application configuration to get the list of workflow branches
+    app_config = config.get_app_config(app_name)
```
benard-g, line 69: If you choose to load the `app_config` from the route function, you could pass it as an argument and not have to retrieve it here.

Also, if you only need the workflow, maybe loading `project_config` is enough (it depends if we want to allow an app to override the default workflow, IMO we can allow it).

nestor_api/lib/workflow.py
```diff
@@ -49,3 +52,92 @@ def get_previous_step(project_config: dict, target: str) -> Optional[str]:
...
+        branch_name for branch_name in workflow_from_conf if branch_name != master_tag
+    ]
+
+    branches = {}
+    status = "success"
+    if len(workflow_branches) != 0:
+        try:
+            # Get user_login linked to the GITHUB_TOKEN
+            user_info = git_provider.get_user_info()
+            print(user_info)
```
benard-g, line 83: Do you really want this print ?

tests/api/api_routes/workflow/test_workflow.py
```diff
@@ -0,0 +1,113 @@
...
+
+        # Assertions
+        self.assertEqual(response.status_code, HTTPStatus.OK)
+        self.assertEqual(
+            data,
+            {
+                "organization": "my-org",
+                "message": "Workflow initialization succeeded",
+                "app": "my-app",
+                "report": {"integration": {"created": [True, True], "protected": [True, True]}},
```
benard-g, line 44: I don't really understand what the two booleans represent, would it make sense to simplify it this way ?
```suggestion
                "report": {"integration": {"created": True, "protected": True}},
```
Which means : "the integration branch has been created and protected"

nestor_api/api/api_routes/workflow/workflow.py
```diff
@@ -0,0 +1,60 @@
+"""Define the workflows routes."""
```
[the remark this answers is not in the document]
chris-verclytte: No I think it will contain all of the workflow controllers. 
But then it should be "Define the workflow controllers."

nestor_api/api/api_routes/workflow/workflow.py
chris-verclytte: But I don't want the app config, I want the project config. The feedback is still valid as there could be some concurrent request for updates but it should not move that often. Anyway, I will create a local copy, for the project config so it will indeed be safer.

nestor_api/api/api_routes/workflow/workflow.py
chris-verclytte: No you're right, it is something I used for debugging before, should have put a "TODO remove" comment ;)

tests/api/api_routes/workflow/test_workflow.py
chris-verclytte: No. First stands for "Was it modified during the operation", second for the current value. 
Could not find a way to document it in a clearer way without a complete documentation

nestor_api/lib/workflow.py
chris-verclytte: I don't choose, this is a port of current implementation where it is working that way. Don't want to be involved in a massive refacto. Let's keep focus on the main objective which is to have something working quickly then improving it later ;)

chris-verclytte: I don't see any difference to have it in argument rather than retrieving it here. This part is the lib responsible for the business logic whereas the route is just a dumb wrapper using it. Seems to me that it does not change anything.

README.md
```diff
@@ -33,3 +33,4 @@ pipenv sync # Add --dev to also install dev dependencies
 |                    `NESTOR_K8S_HTTP_PROXY` |                        |            | The kubernetes HTTP_PROXY                                   |
 |                  `NESTOR_K8S_SERVICE_PORT` | `8080`                 |            | The port on which the k8s services will be exposed          |
 |               `NESTOR_K8S_TEMPLATE_FOLDER` | `templates`            |            | The subfolder in which the k8s templates are stored         |
+|                    `NESTOR_GIT_MASTER_TAG` | `master`               |            | The tag used to define the master branch                    |
```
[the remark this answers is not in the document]
chris-verclytte: I don't think that default branch would be a better naming, master is not a "default" branch IMO, is it ?

nestor_api/api/api_routes/workflow/workflow.py
```diff
@@ -0,0 +1,72 @@
...
+            raise ValueError(f"Unexpected status: '{report_status}'")
+
+        # Clean the temporary directory
+        try:
+            io_lib.remove(config_dir)
+        # pylint: disable=broad-except
+        except Exception as err:
+            Logger.warn(
+                {"config_dir": config_dir, "err": err}, "Failed to clean temporary config dir"
+            )
```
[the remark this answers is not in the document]
chris-verclytte: Yes I put it here to avoid having this if nesting

README.md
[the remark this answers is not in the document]
benard-g: I suggested this because Github refers to the master branch as the `default branch`.
But no big deal.

chris-verclytte:
Aaaah, then could make sense. Even if I am not aware this is common with other providers 🤔

README.md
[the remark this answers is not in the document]
chris-verclytte: Aaaah, then could make sense. Even if I am not aware this is common with other providers 🤔

nestor_api/lib/workflow.py
```diff
@@ -1,10 +1,42 @@
...
+
+
+class BranchReport(TypedDict, total=False):
+    """Branch report dictionary"""
+
+    created: CreationStatus
+    protected: ProtectionStatus
+
+
+# BranchReport = Union[, Dict[str, ]]
```
benard-g, line 38: Forgotten comment ?

PR #64, opened 2020-07-27
Add nestor configuration example and documentation

maximejimenez:
Add an example for configuration and documentation.
We will see to extract this later on in order to provide a **forkable** repository with the **branches matching the defined workflows in the example**.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1,0 +1,34 @@
+# Nestor configuration
+
+## Context
+
+Nestor configuration contains configuration of how apps are configured and deployed.
+
+## How to use it
+
+Nestor configuration's should be in a dedicated repository, so that you can keep track of changes.
+The repository should have one branch for each step of the workflow defined in the `project.yaml`.
+
+## Details
+
+### Project
+
+The `project.yaml` contains global configuration.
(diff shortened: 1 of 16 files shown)


docs/configuration.md
```diff
@@ -0,0 +1,34 @@
...
+
+### Project
+
+The `project.yaml` contains global configuration.
+You can find default values in this file but also global configuration (e.g. docker registry, workflow's steps).
+
+### Application
+
+Applications' configurations can be found under `apps` folder.
+These configuration will be merged with the `project.yaml` configuration.
```
Kenny407, line 22:
```suggestion
This configuration will be merged with the `project.yaml` configuration.
```

docs/configuration.md
```diff
@@ -0,0 +1,34 @@
...
+If values are defined in both project configuration and application configuration, application will overwrite project.
```
Kenny407, line 23: Maybe we can rewrite this part to explain just that application values have higher priority than the default project values?

docs/configuration.md
Kenny407: Or `these configurations`? Not sure what suits better

PR #67, opened 2020-07-28
Add a template loader for the k8s/builder

benard-g:
Add pybars3 dependency to resolve handlebar templates.

The templates are not used yet but will be in a next PR on kubernetes' `builders`.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -19,6 +19,7 @@ PyYAML = "==5.3.1"
 semver = "==2.10.2"
 jsonschema = "==3.2.0"
 pygithub = "==1.51"
+pybars3 = "*"
 
 [requires]
 python_version = "3.7"
--- a/Pipfile
+++ b/Pipfile
@@ -19,7 +19,7 @@ PyYAML = "==5.3.1"
 semver = "==2.10.2"
 jsonschema = "==3.2.0"
 pygithub = "==1.51"
-pybars3 = "*"
+pybars3 = "0.9.7"
 
 [requires]
 python_version = "3.7"
--- a/README.md
+++ b/README.md
@@ -31,3 +31,4 @@ pipenv sync # Add --dev to also install dev dependencies
 |              `NESTOR_REPLICAS_DEFAULT_MAX` | `10`                   | `replicas` | Default maximum number of replicas                          |
 | `NESTOR_REPLICAS_DEFAULT_TARGET_CPU_USAGE` | `75`                   | `%`        | Default target cpu usage that will trigger an autoscaling   |
 |                    `NESTOR_K8S_HTTP_PROXY` |                        |            | The kubernetes HTTP_PROXY                                   |
+|               `NESTOR_K8S_TEMPLATE_FOLDER` | `templates`            |            | The subfolder in which the "build" templates will be loaded |
--- a/README.md
+++ b/README.md
@@ -31,4 +31,4 @@ pipenv sync # Add --dev to also install dev dependencies
 |              `NESTOR_REPLICAS_DEFAULT_MAX` | `10`                   | `replicas` | Default maximum number of replicas                          |
 | `NESTOR_REPLICAS_DEFAULT_TARGET_CPU_USAGE` | `75`                   | `%`        | Default target cpu usage that will trigger an autoscaling   |
 |                    `NESTOR_K8S_HTTP_PROXY` |                        |            | The kubernetes HTTP_PROXY                                   |
-|               `NESTOR_K8S_TEMPLATE_FOLDER` | `templates`            |            | The subfolder in which the "build" templates will be loaded |
+|               `NESTOR_K8S_TEMPLATE_FOLDER` | `templates`            |            | The subfolder in which the k8s templates are stored         |
--- a/nestor_api/config/k8s.py
+++ b/nestor_api/config/k8s.py
@@ -12,5 +12,5 @@ class K8sConfiguration:
 
     @staticmethod
     def get_templates_dir() -> str:
-        """Returns the name of the config sub-folder in which to look for templates."""
+        """Returns the subfolder in which the k8s templates are stored."""
         return os.getenv("NESTOR_K8S_TEMPLATE_FOLDER", "templates")
--- a/nestor_api/config/k8s.py
+++ b/nestor_api/config/k8s.py
@@ -2,7 +2,6 @@
 import os
 
 
-# pylint: disable=too-few-public-methods
 class K8sConfiguration:
     """k8s configuration"""
 
@@ -10,3 +9,8 @@ class K8sConfiguration:
     def get_http_proxy() -> str:
         """Returns the k8s http proxy"""
         return os.environ["NESTOR_K8S_HTTP_PROXY"]
+
+    @staticmethod
+    def get_templates_dir() -> str:
+        """Returns the name of the config sub-folder in which to look for templates."""
+        return os.getenv("NESTOR_K8S_TEMPLATE_FOLDER", "templates")
--- a/nestor_api/lib/k8s/builders.py
+++ b/nestor_api/lib/k8s/builders.py
@@ -1,11 +1,38 @@
 """k8s deployment file builders"""
 
+import os
 from typing import Optional
+
+from pybars import Compiler  # type: ignore
 
 from nestor_api.config.probes import ProbesDefaultConfiguration
 from nestor_api.config.replicas import ReplicasDefaultConfiguration
 import nestor_api.lib.docker as docker
+import nestor_api.lib.io as io
 import yaml_lib
+
+TEMPLATES = [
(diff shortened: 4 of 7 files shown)


nestor_api/config/k8s.py
```diff
@@ -2,11 +2,15 @@
...
 
     @staticmethod
     def get_http_proxy() -> str:
         """Returns the k8s http proxy"""
         return os.environ["NESTOR_K8S_HTTP_PROXY"]
+
+    @staticmethod
+    def get_templates_dir() -> str:
+        """Returns the name of the config sub-folder in which to look for templates."""
+        return os.getenv("NESTOR_K8S_TEMPLATE_FOLDER", "templates")
```
maximejimenez, line 16: IMO this can be resolved using NESTOR_CONFIG_PATH concatenated with `templates`

Pipfile
```diff
@@ -19,6 +19,7 @@ PyYAML = "==5.3.1"
 semver = "==2.10.2"
 jsonschema = "==3.2.0"
 pygithub = "==1.51"
+pybars3 = "*"
```
maximejimenez, line 22: I would freeze the version here
Don't forget to update again `Pipfile.lock` after changing this

nestor_api/lib/k8s/builders.py
```diff
@@ -1,12 +1,39 @@
...
+TEMPLATES = [
+    "deployment",
+    "hpa",
+    "anti-affinity-node",
+    "anti-affinity-zone",
+    "service",
+    "namespace",
+    "cronjob",
+    "job",
+]
```
maximejimenez, line 23:
```suggestion
TEMPLATES = [
    "deployment",
    "hpa",
    "anti-affinity-node",
    "anti-affinity-zone",
    "service",
    "namespace",
    "cronjob",
]
```

tests/lib/k8s/test_builders.py
```diff
@@ -42,6 +42,22 @@ def check_namespace(parameters):
...
+        templates = k8s_builders.load_templates("/path", ["template_a", "template_b"])
+
+        self.assertEqual(list(templates.keys()), ["template_a", "template_b"])
+        io_mock.read.assert_has_calls(
+            [call("/path/template_a.yaml"), call("/path/template_b.yaml")]
+        )
+        compile_mock.assert_has_calls(
+            [call("file: /path/template_a.yaml"), call("file: /path/template_b.yaml")]
+        )
+
```
maximejimenez, line 60: We could also try to add one more test where we use a template from fixtures and we try to use it after compilation to ensure it actually fits our needs and lock this behavior

nestor_api/config/k8s.py
benard-g: I plan to use it this way :
```python
from nestor_api.config.k8s import K8sConfiguration
import nestor_api.lib.config as config

config_dir = config.create_temporary_config_copy()
templates_path = os.path.join(config_dir, K8sConfiguration.get_templates_dir())
```

benard-g: (This makes me realize that I forgot to reference the variable in the README)

nestor_api/lib/k8s/builders.py
benard-g: I will actually need the `job` to build the cronjob.

nestor_api/lib/k8s/builders.py
```diff
@@ -1,12 +1,39 @@
...
+
+
+def load_templates(templates_path: str, template_names: list) -> dict:
+    """Load the templates from the configuration path"""
+    template_compiler = Compiler()
+
+    def _load(template):
```
chris-verclytte, line 30: Why defining this function internally rather than in the same scope ?

tests/lib/k8s/test_builders.py
```diff
@@ -42,6 +42,25 @@ def check_namespace(parameters):
             "namespace": check_namespace,
         }
 
+    @patch("nestor_api.lib.k8s.builders.io", autospec=True)
+    def test_load_templates(self, io_mock):
+        """Should load the templates and correctly substitute `{{variable}}`"""
+
+        def _read_mock(file_name):
```
chris-verclytte, line 49: Nit picking: Maybe a lambda would improve the readability here

nestor_api/config/k8s.py
```diff
@@ -2,11 +2,15 @@
...
     """k8s configuration"""
 
     @staticmethod
     def get_http_proxy() -> str:
         """Returns the k8s http proxy"""
         return os.environ["NESTOR_K8S_HTTP_PROXY"]
+
+    @staticmethod
+    def get_templates_dir() -> str:
+        """Returns the name of the config sub-folder in which to look for templates."""
```
chris-verclytte, line 15: I am pretty sure this sentence is not correct english but I am not sure about how it should be actually 😅
Maybe:
`Returns the config sub-folder's name in which looking for templates.`

nestor_api/lib/k8s/builders.py
benard-g: It was mostly to take advantage of the variables scoping here but I can write it this way if you prefer :
```python
def _load(templates_path, template_compiler, template_name):
    template_path = os.path.join(templates_path, f"{template_name}.yaml")
    file_content = io.read(template_path)
    return template_compiler.compile(file_content)


def load_templates(templates_path: str, template_names: list) -> dict:
    """Load the templates from the configuration path"""
    template_compiler = Compiler()
    return {
        template: _load(templates_path, template_compiler, template) for template in template_names
    }
```
I personally prefer the first version 😛

tests/lib/k8s/test_builders.py
benard-g: I tried and it looked indeed better 👍

nestor_api/config/k8s.py
benard-g: What about ?
```
Returns the name of the folder holding the templates. It will be resolved under the `config` path.
```

nestor_api/lib/k8s/builders.py
chris-verclytte: For me, nesting is more to be used when your child function does not make sense without its parent which IMO is not the case here. But, maybe I am wrong, don't hesitate if you have some sources saying the inverse. I think that's not a really important feedback so don't bother too much ;)

nestor_api/config/k8s.py
chris-verclytte: Ahah actually I just red back the name of the function and it is definitely more explicit than any of those description we suggested 😅

README.md
```diff
@@ -31,3 +31,4 @@ pipenv sync # Add --dev to also install dev dependencies
 |              `NESTOR_REPLICAS_DEFAULT_MAX` | `10`                   | `replicas` | Default maximum number of replicas                          |
 | `NESTOR_REPLICAS_DEFAULT_TARGET_CPU_USAGE` | `75`                   | `%`        | Default target cpu usage that will trigger an autoscaling   |
 |                    `NESTOR_K8S_HTTP_PROXY` |                        |            | The kubernetes HTTP_PROXY                                   |
+|               `NESTOR_K8S_TEMPLATE_FOLDER` | `templates`            |            | The subfolder in which the "build" templates will be loaded |
```
maximejimenez, line 34:
```suggestion
|               `NESTOR_K8S_TEMPLATE_FOLDER` | `templates`            |            | The subfolder in which the k8s templates are located |
```

maximejimenez:
```suggestion
|               `NESTOR_K8S_TEMPLATE_FOLDER` | `templates`            |            | The subfolder in which the k8s templates are stored |
```

nestor_api/config/k8s.py
chris-verclytte: Then "Get directory containing Kubernetes templates" would be probably the clearest solution :)

nestor_api/lib/k8s/builders.py
```diff
@@ -1,12 +1,42 @@
...
+    "anti-affinity-node",
+    "anti-affinity-zone",
+    "service",
+    "namespace",
+    "cronjob",
+    "job",
+]
+
+
+def _load_template(templates_path: str, template_compiler: Compiler, template_name: str):
```
chris-verclytte, line 26: Docstring is missing here

tests/lib/k8s/test_builders.py
```diff
@@ -42,6 +42,24 @@ def check_namespace(parameters):
             "namespace": check_namespace,
         }
 
+    @patch("nestor_api.lib.k8s.builders.io", autospec=True)
+    def test_load_templates(self, io_mock):
+        """Should load the templates and correctly substitute `{{variable}}`"""
```
chris-verclytte, line 47: Following PEP, docstring should end with a period (I know, it is quite strict...)

PR #68, opened 2020-07-29
Add new library functions needed for the deployment_api

benard-g:
This PR adds a few functionalities to the existing `lib/` modules that will be used by the deployment_api in a later PR.

Hint: I split each new function commit-by-commit to improve the readability 🙂

### DIFF
--- a/nestor_api/lib/config.py
+++ b/nestor_api/lib/config.py
@@ -68,6 +68,18 @@ def get_project_config(config_path: str = Configuration.get_config_path()) -> di
     return _resolve_variables_deep(project_config)
 
 
+def get_deployments_configs(project_config: dict) -> list:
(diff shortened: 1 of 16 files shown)


nestor_api/lib/config.py
```diff
@@ -68,6 +68,18 @@ def get_project_config(config_path: str = Configuration.get_config_path()) -> di
     return _resolve_variables_deep(project_config)
 
 
+def get_deployments_configs(project_config: dict) -> list:
```
maximejimenez, line 71:
```suggestion
def get_deployments(project_config: dict) -> list:
```

Otherwise you will end-up with `config. get_deployments_configs()` calls which are a bit strange IMO

PR #69, opened 2020-07-30
Add setters in k8s/builder

benard-g:
This PR includes all the setters that will then be used when configuring the deployment resources.
I choose to create a dedicated PR to keep things simple.

### DIFF
--- a/nestor_api/lib/k8s/builders.py
+++ b/nestor_api/lib/k8s/builders.py
@@ -1,7 +1,7 @@
 """k8s deployment file builders"""
 
 import os
-from typing import Optional
+from typing import Optional, Tuple
 
 from pybars import Compiler  # type: ignore
 
@@ -9,7 +9,12 @@ from nestor_api.config.probes import ProbesDefaultConfiguration
 from nestor_api.config.replicas import ReplicasDefaultConfiguration
 import nestor_api.lib.docker as docker
 import nestor_api.lib.io as io
+import nestor_api.utils.dict as dict_utils
+import nestor_api.utils.list as list_utils
 import yaml_lib
+
+SERVICE_PORT = 8080
+
 
 TEMPLATES = [
     "deployment",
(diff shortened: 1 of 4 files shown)


nestor_api/lib/k8s/builders.py
```diff
@@ -1,16 +1,21 @@
...
 
 from nestor_api.config.probes import ProbesDefaultConfiguration
 from nestor_api.config.replicas import ReplicasDefaultConfiguration
 import nestor_api.lib.docker as docker
 import nestor_api.lib.io as io
+import nestor_api.utils.dict as dict_utils
+import nestor_api.utils.list as list_utils
 import yaml_lib
 
+SERVICE_PORT = 8080
```
maximejimenez, line 16: I think this should be configurable with a env variable

tests/lib/k8s/test_builders.py
```diff
@@ -176,17 +176,50 @@ def test_get_anti_affinity_zone_enabled_for_process(self, parse_yaml_mock):
...
+            "registry": {"platform": "docker-hub", "id": "id-1"},
+            "docker": {
+                "registries": {
+                    "docker-hub": [
+                        {"id": "id-1", "organization": "my-organization-1"},
+                        {"id": "id-2", "organization": "my-organization-2"},
+                    ],
+                },
+            },
+        }
```
maximejimenez, line 190: I would personally not have this over-complicated
There is no real reason to have several registries ATM

tests/lib/k8s/test_builders.py
```diff
@@ -176,17 +176,50 @@ def test_get_anti_affinity_zone_enabled_for_process(self, parse_yaml_mock):
...
         image_url = k8s_builders.get_image_name(
             app_config, {"branch": "feature/api", "tag": "1.0.0-sha-a2b3c4"}
         )
-        self.assertEqual(image_url, "my-organization/app:feature/api")
+        self.assertEqual(image_url, "my-organization-1/app:feature/api")
```
maximejimenez, line 194: It does not make sense to have this return the first and not the second one IMO, that's why we should keep it simple

nestor_api/lib/k8s/builders.py
```diff
@@ -39,53 +44,70 @@ def load_templates(templates_path: str, template_names: list) -> dict:
...
+def get_image_name(deployment_config: dict, options: dict) -> str:
     """Returns the definitive image name."""
+    image_name = deployment_config.get("docker", {}).get("image_name") or deployment_config["app"]
+
+    registry_ref = deployment_config["registry"]
+    registries = deployment_config["docker"]["registries"][registry_ref["platform"]]
+    registry = list_utils.find(registries, lambda reg: reg["id"] == registry_ref["id"])
+    if registry is None:
+        raise ValueError(f'No registry matching "{registry_ref["id"]}"')
+
```
maximejimenez, line 104: I believe we are kind of over-complicating things up here by changing this

PR #76, opened 2020-08-04
Add k8s.deployment lib with a deploy_app 

benard-g:


nestor_api/lib/k8s/builders.py
```diff
@@ -36,6 +34,30 @@ def build_yaml(deployment_config: dict, templates_path: str, tag_to_deploy: str)
     return "\n".join(yaml_sections)
 
 
+def build_ingress_yaml(deployment_config: dict, templates: dict) -> str:
+    """Builds the k8s yaml corresponding to the provided configuration."""
+    app_name, sanitized_process_name, metadata_name = get_sanitized_names(deployment_config, "web")
```
maximejimenez, line 39:
```suggestion
    app_name, sanitized_process_name, metadata_name = get_sanitized_names(deployment_config, WEB_PROCESS_NAME)
```

Maybe use the defined `WEB_PROCESS_NAME = "web"`

tests/lib/k8s/test_builders.py
```diff
@@ -1300,3 +1311,30 @@ def test_get_sections_for_cronjob_with_namespace(
...
+
+    def test_build_ingress_yaml(self):
+        # Mocks
+        deployment_config = {
+            "app": "my-app",
+            "domain": "my-app.com",
+            "domain_prefix": "-staging",
+        }
+        templates = {
+            "ingress-app": self._create_template_validator(
```
maximejimenez, line 1323: Maybe it would be good to have all expectation use this regarding templates and remove `check_hpa`, `check_namespace`, etc.

nestor_api/lib/k8s/__init__.py
```diff
@@ -0,0 +1,59 @@
+"""Kubernetes library."""
```
maximejimenez, line 1: I would say that it is a bit strange to define all of this in the `__init__.py`

tests/lib/k8s/__init__.py
```diff
@@ -0,0 +1,146 @@
+from unittest import TestCase
```
maximejimenez, line 1: Same comment regarding `__init__.py`

nestor_api/lib/k8s/builders.py
benard-g: oops 😆

benard-g: I will pass it as an argument in `build_ingress_yaml` to prevent introducing circular dependencies between the two modules.

PR #82, opened 2020-08-05
Documentation of build API

maximejimenez:
Add documentation for build API.
Explain in the `getting started` what should be configured and why.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1,0 +1,10 @@
+# API
+
+## Builds
+
+### POST `/api/builds/:app`
+
+<a name="api-build-api"></a>
+
+Builds the docker image of an application from the first step defined in the workflow with a unique
+tag and uploads it to the configured Docker registry.
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -8,3 +8,12 @@
 Create an account with sufficient permissions to be able to merge on protected branches using the workflow features.
 
 **Supported host:** GitHub
+
+## Configuration
+
+### Build
+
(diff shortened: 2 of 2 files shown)


docs/api.md
```diff
@@ -0,0 +1,10 @@
+# API
+
+## Builds
+
+### POST `/api/builds/:app`
+
+<a name="api-build-api"></a>
+
+Builds the docker image of an application from the master branch with a unique
+tag and upload it to the configured Docker registry.
```
benard-g, line 10:
```suggestion
Builds the docker image of an application from the `master` branch with a unique
tag and uploads it to the configured Docker registry.
```
Also, I believe it is not especially `master` but rather "the first step defined in the workflow".

docs/getting_started.md
```diff
@@ -8,3 +8,12 @@
...
 **Supported host:** GitHub
+
+## Configuration
+
+### Build
+
+You should create a hook that is called after merge on the first step of your workflow (e.g. master).
+This will create the docker image by calling the [build API](./api.md#api-build-api).
+The created image docker will be tagged with a reference to the version of your application and the commit hash.
+**This will guarantee that you are using the same version for all environment.**
```
benard-g, line 19: What about ?
```suggestion
You should create a hook triggered every time new commits are merged on the first step of your workflow (e.g. master) calling [Nestor's build API](./api.md#api-build-api).
This will create a new docker image corresponding to this new version. Nestor will tag this image with a reference to the version of your application and the commit hash before adding it to your Docker registry.
**This will guarantee that you are using the same version for all environment.**
```
You could also try to add some spacing because everything is rendered on a single line.

PR #88, opened 2020-08-07
Add git rebase onto option

chris-verclytte:
Will be useful for advance_workflow.
Externalize it to reduce pull request size.

### DIFF
--- a/nestor_api/lib/git.py
+++ b/nestor_api/lib/git.py
@@ -72,9 +72,11 @@ def push(repository_dir: str, branch_name: str = "HEAD") -> None:
     io.execute(f"git push origin {branch_name} --tags --follow-tags", repository_dir)
 
 
-def rebase(repository_dir: str, branch_name: str) -> None:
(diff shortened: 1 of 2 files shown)


nestor_api/lib/git.py
```diff
@@ -72,9 +72,11 @@ def push(repository_dir: str, branch_name: str = "HEAD") -> None:
     io.execute(f"git push origin {branch_name} --tags --follow-tags", repository_dir)
 
 
-def rebase(repository_dir: str, branch_name: str) -> None:
+def rebase(repository_dir: str, branch_name: str, *, onto: str = None) -> None:
```
Kenny407, line 75: Cool feature,  didn't know this was how it was enforced 👍

chris-verclytte: @Kenny407 Yep, you can dig the subject by reading [this section](https://docs.python.org/3/tutorial/controlflow.html#more-on-defining-functions) of the Python tutorial which explains well all of the possible function arguments 🙂

